PR #14, opened 2025-08-22
feat: standby testing

### LINKED ISSUE #12: standby tests
flow:
- build `0.99` version
- create new task `test-0.99.X`
- call the task's standby url `https://ondrejklinovsky--test-0-99-X.apify.actor` to start the correct version
- delete the task after all tests are finished

oklinov:
closes #12

### DIFF
--- a/lib/lib.ts
+++ b/lib/lib.ts
@@ -1,4 +1,4 @@
-import { ApifyClient } from 'apify-client';
+import { Actor, ActorRun, ActorRunListItem, ActorStandby, ApifyClient, Task } from 'apify-client';
 import {
     describe as vitestDescribe,
     ExpectStatic,
@@ -11,7 +11,7 @@ import {
 import type { ActorBuild, RunOptions } from './types';
 import { RunTestResult } from './run-test-result.js';
 import { extendExpect } from './extend-expect.js';
-import { getActorPrefilledInput } from './utils.js';
+import { getActorPrefilledInput, sleep } from './utils.js';
 
 const ACTOR_BUILDS = 'ACTOR_BUILDS';
 let actorBuilds: ActorBuild[] = [];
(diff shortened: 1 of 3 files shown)


oklinov:
@metalwarrior665 @Real1tyy could you please take a look? I'm thinking whether the API should be different as it doesn't allow us to send multiple requests to a single standby run :thinking: wdyt?

lib/lib.ts
```diff
@@ -80,6 +80,47 @@ export const testActor = <T>(
     });
 };
 
+/**
+ * This wrapper creates a new task with specific `build` of the standby actor and provides
+ * `callStandby` function, which calls the task's `standbyUrl`.
```
metalwarrior665, line 85: Add that using task is just current shortcoming of standby feature but ideally we would use Actor directly

lib/lib.ts
```diff
@@ -97,7 +138,91 @@ export const testTestActor = <T>(
...
+    const { isEnabled, ...defaultActorStandby } = actorInfo.actorStandby;
+    const build = buildNumber ?? defaultActorStandby.build;
+
+    const actorStandbyOptions: ActorStandby = {
+        ...defaultActorStandby,
+        build,
+    }
+
+    try {
+        const newTask = await apifyClient.tasks().create({
```
metalwarrior665, line 201: You are missing `name` which is the technical name. It should also contain the tech name of the Actor + build so it doesn't clash with other tasks.

lib/lib.ts
```diff
@@ -80,6 +80,47 @@ export const testActor = <T>(
     });
 };
 
+/**
+ * This wrapper creates a new task with specific `build` of the standby actor and provides
+ * `callStandby` function, which calls the task's `standbyUrl`.
+ */
```
metalwarrior665, line 88: The description should probably also mention that `testStandbyActor` should be only called once per Actor since now it doesn't fetch the existing task. I'm fine to keep this restriction as it is but let's document it.

Another option would be to have each `testStandbyActor` some unique identifier that would be put to the task name

metalwarrior665 (approved):
Looks good

lib/lib.ts
oklinov: you can call it as many times as you want...every `testStandbyActor`  creates it's own task and after it's done the task is deleted

lib/lib.ts
oklinov: I omitted `name` on purpose so that the platform creates the technical name on its own...this way I don't have to deal with conflicts

Real1tyy (approved):

PR #25, opened 2025-09-19
fix: parse JSON in a specific function

### LINKED ISSUE #20: commit message containing double quotes breaks git -> json conversion
The following command:
```sh
git log -1 --pretty=format:'{"sha":"%H","author":"%aN<%aE>","date":"%aD","message":"%s"}' origin/test/commit-with-double-quotes
```
generates the following output:
returns:
```json
{
    "sha":"d4afa862921dec6963e7d6895538219a7525036e",
    "author":"oklinov<oklinov@tutanota.com>",
    "date":"Mon, 8 Sep 2025 09:38:49 +0200",
    "message":"test: commit with double quotes: ""
}
```

which is an invalid json. We need to somehow escape the quotes.

ruocco-l:
Closes #20 

---

@kazadoiyul , I'll merge the changes from the comments by @oklinov and then set this to review but if you want you can already have a look

### DIFF
--- a/bin/git.ts
+++ b/bin/git.ts
@@ -1,11 +1,12 @@
 import { Commit, Config } from './types';
-import { spawnCommandInGhWorkspace } from './utils.js';
+import { parseCommitAsJson, spawnCommandInGhWorkspace } from './utils.js';
 
-const GIT_LOG_FORMAT = `{"sha":"%H","author":"%aN<%aE>","date":"%aD","message":"%s"}`;
+const GIT_LOG_FORMAT = `%H»¦«%aN<%aE>»¦«%aD»¦«%s`;
 
 export const getBranchLastCommit = (branch: string): Commit => {
-    const lastCommit = JSON.parse(spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${branch}`)) as Commit;
-    return lastCommit;
+    const stringFormattedCommit = spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${branch}`);
+    const jsonParsedCommit = parseCommitAsJson(stringFormattedCommit);
+    return jsonParsedCommit;
 };
 
 export const getChangedFiles = (commits: Commit[]) => {
@@ -37,12 +38,12 @@ export const getCommits = ({
     const base = getBaseCommit(sourceBranchCommit, targetBranchCommit);
     const commitsStrings = spawnCommandInGhWorkspace(`git log --pretty=format:'${GIT_LOG_FORMAT}' ${base.sha}..${sourceBranchCommit.sha}`)
         .split('\n');
-    const commits = JSON.parse(`[${commitsStrings.join(',')}]`) as Commit[];
-    commits.reverse();
+    const commitsJsons = commitsStrings.map((commitString) => parseCommitAsJson(commitString)) as Commit[];
+    commitsJsons.reverse();
     const baseCommitIndex = baseCommit
-        ? commits.findIndex(({ sha }) => sha === baseCommit.sha)
+        ? commitsJsons.findIndex(({ sha }) => sha === baseCommit.sha)
         : -1;
-    return commits.slice(baseCommitIndex + 1);
+    return commitsJsons.slice(baseCommitIndex + 1);
 };
 
 export const getBaseCommit = (target: Commit, source: Commit): Commit => {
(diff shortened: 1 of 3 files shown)


bin/utils.ts
```diff
@@ -86,6 +87,19 @@ export const getHeadCommitSha = (githubEvent: GitHubEvent) => {
         : githubEvent.head_commit.id;
 };
 
+export const parseCommitAsJson = (commitString: string): Commit => {
```
oklinov: I'd move this to `git.ts` as it's related only to git

bin/git.ts
```diff
@@ -1,11 +1,12 @@
 import { Commit, Config } from './types';
-import { spawnCommandInGhWorkspace } from './utils.js';
+import { parseCommitAsJson, spawnCommandInGhWorkspace } from './utils.js';
 
-const GIT_LOG_FORMAT = `{"sha":"%H","author":"%aN<%aE>","date":"%aD","message":"%s"}`;
+const GIT_LOG_FORMAT = `%H»¦«%aN<%aE>»¦«%aD»¦«%s`;
```
oklinov: I'd create a const `GIT_FORMAT_SEPARATOR = '»¦«'` so it's defined only once and then define `GIT_LOG_FORMAT` as:
```suggestion
const GIT_LOG_FORMAT = ['%H', '%aN<%aE>', '%aD', '%s'].join(GIT_FORMAT_SEPARATOR);
```

though I might be less readible this way...I'll leave it up to you :slightly_smiling_face:

bin/git.ts
```diff
@@ -37,12 +38,12 @@ export const getCommits = ({
     const base = getBaseCommit(sourceBranchCommit, targetBranchCommit);
     const commitsStrings = spawnCommandInGhWorkspace(`git log --pretty=format:'${GIT_LOG_FORMAT}' ${base.sha}..${sourceBranchCommit.sha}`)
         .split('\n');
-    const commits = JSON.parse(`[${commitsStrings.join(',')}]`) as Commit[];
-    commits.reverse();
+    const commitsJsons = commitsStrings.map((commitString) => parseCommitAsJson(commitString)) as Commit[];
```
oklinov: I'd call it just `commits` and `parseCommit`. When I see `somethingJson` it may indicate that it's json formatted string instead of plain JS object
```suggestion
    const commits = commitsStrings.map((commitString) => parseCommit(commitString)) as Commit[];
```

bin/git.ts
```diff
@@ -1,11 +1,12 @@
...
+import { parseCommitAsJson, spawnCommandInGhWorkspace } from './utils.js';
 
-const GIT_LOG_FORMAT = `{"sha":"%H","author":"%aN<%aE>","date":"%aD","message":"%s"}`;
+const GIT_LOG_FORMAT = `%H»¦«%aN<%aE>»¦«%aD»¦«%s`;
 
 export const getBranchLastCommit = (branch: string): Commit => {
-    const lastCommit = JSON.parse(spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${branch}`)) as Commit;
-    return lastCommit;
+    const stringFormattedCommit = spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${branch}`);
+    const jsonParsedCommit = parseCommitAsJson(stringFormattedCommit);
```
oklinov: same as bellow :arrow_down: - I wouldn't use 'json' in vars names:
```suggestion
    const lastCommit = parseCommit(stringFormattedCommit);
```

bin/git.ts
```diff
@@ -52,5 +53,7 @@ export const getBaseCommit = (target: Commit, source: Commit): Commit => {
 };
 
 export const getCommitInfo = (commitSha: string): Commit => {
-    return JSON.parse(spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${commitSha}`));
+    const stringParsedCommit = spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${commitSha}`);
+    const jsonParsedCommit = parseCommitAsJson(stringParsedCommit);
+    return jsonParsedCommit;
```
oklinov:
```suggestion
    const stringCommit = spawnCommandInGhWorkspace(`git log -1 --pretty=format:'${GIT_LOG_FORMAT}' ${commitSha}`);
    const commit = parseCommit(stringParsedCommit);
    return commit;
```

bin/utils.ts
```diff
@@ -86,6 +87,19 @@ export const getHeadCommitSha = (githubEvent: GitHubEvent) => {
...
+    const [sha, author, date, message] = commitString.split('»¦«');
+    if (!sha || !author || !date || !message) {
```
oklinov: According to chatgpt, some of these may be empty:

> ✅ So in short:
> Always non-empty: %H, %aD
> Can be empty: %aN, %aE, %s

so maybe it'd be safer to check the length:
```ts
const splits = commitString.split('»¦«');
if (splits.length !== 4) { ... }
const [sha, author, date, message] = splits;

oklinov (commented):
thanks, overall good. just found couple of details :slightly_smiling_face: 

could you also write some unit tests for the commit parsing?

bin/git.ts
ruocco-l: Smart

bin/utils.ts
ruocco-l: Yeah, I wasn't confident on this check right from the start. But maybe we want to allow the message to be empty?

oklinov: yeah, I would allow it. Comparing `splits.length !== 4` will work also with an empty commit message. `.. || !message` would fail.

oklinov (approved):

PR #26, opened 2025-09-19
feat: add workspace option and make GITHUB_WORKSPACE env variable optional

### LINKED ISSUE #21: don't require GITHUB_WORKSPACE env variable
`GITHUB_WORKSPACE` is currently required as it tells the cli where the repository is located. We should make it optional and add another ways to specify path to repository:
- cli argument `--workspace <path>` - this should take precedence over the `GITHUB_WORKSPACE`
- if neither cli argument nor  `GITHUB_WORKSPACE` is specified, we should use current working directory

ruocco-l:
Closes #21 

~~Not the prettiest code to just drill the option all the way through but it should work, if you have any remarks on it please let me know.~~

@kazadoiyul 

(@oklinov I'm not sure about `process.cwd()`, educate me if, for some reasons, it's not correct in this context)

Edit: tried new approach using middlewares. If you are ok with it we can also refactor a bit (to avoid having a gigantic `utils.ts`)

### DIFF
--- a/bin/utils.ts
+++ b/bin/utils.ts
@@ -5,9 +5,9 @@ import type {
     GitHubEvent,
 } from './types.js';
 
-export const spawnCommandInGhWorkspace = (command: string, args: string[] = []) => {
+export const spawnCommandInGhWorkspace = ({ command, workspace, args = [] }: { command: string, workspace?: string, args?: string[] }) => {
     console.error(command, args.join(' '));
-    const ghWorkspace = getEnvVar('GITHUB_WORKSPACE');
+    const ghWorkspace = workspace ?? getEnvVar('GITHUB_WORKSPACE', process.cwd());
     const commandInGhWorkspace = `cd ${ghWorkspace}; ${command}`;
     const commandResult = spawnSync(commandInGhWorkspace, args, { shell: true, maxBuffer: 100 * 1024 * 1024 });
 
--- a/bin/utils.ts
+++ b/bin/utils.ts
@@ -7,17 +7,16 @@ import type {
 
 export const spawnCommandInGhWorkspace = (command: string, args: string[] = []) => {
     console.error(command, args.join(' '));
-    const commandInGhWorkspace = `cd ${process.cwd()}; ${command}`;
-    const commandResult = spawnSync(commandInGhWorkspace, args, { shell: true, maxBuffer: 100 * 1024 * 1024 });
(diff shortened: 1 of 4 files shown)


oklinov (commented):
thanks for the PR.

I think using `process.cwd()` is the right way :+1: 

Though I'm not a fan of the `workspace` prop drilling :thinking: What if we used `process.chdir` to set CWD before calling `getCommits`, `getChangedFiles` etc? This way we wouldn't have to worry about CWD in `spawnCommandInGhWorkspace` and therefore we woudn't have to pass the `workspace` down the flow. wdyt?

bin/utils.ts
```diff
@@ -7,8 +7,7 @@ import type {
 
 export const spawnCommandInGhWorkspace = (command: string, args: string[] = []) => {
     console.error(command, args.join(' '));
-    const ghWorkspace = getEnvVar('GITHUB_WORKSPACE');
-    const commandInGhWorkspace = `cd ${ghWorkspace}; ${command}`;
+    const commandInGhWorkspace = `cd ${process.cwd()}; ${command}`;
```
oklinov: `cd ${process.cwd()}` doesn't do anything (you're changing directory to current working directory), so you can just pass `command` directly yo `spawnSync`

bin/utils.ts
```diff
@@ -47,17 +46,16 @@ export const getRunUrlKvsKey = (runnerName: string) => {
  * This works locally if checkoutRepoLocally is called first
  */
 export const getRepoActors = async (): Promise<ActorConfig[]> => {
-    const ghWorkspace = getEnvVar('GITHUB_WORKSPACE');
     let actorDirs: string[];
     try {
-        actorDirs = (await fs.readdir(`${ghWorkspace}/actors`)).map((dir) => `actors/${dir}`);
+        actorDirs = (await fs.readdir(`${process.cwd()}/actors`)).map((dir) => `actors/${dir}`);
```
oklinov: you could also use a relative path, IMO it's a bit more readable. I'll leave up to you :slightly_smiling_face: 
```suggestion
        actorDirs = (await fs.readdir('./actors')).map((dir) => `actors/${dir}`);
```
same applies to line 58

oklinov (changes requested):
nice work with using middleware. I left two last comments

bin/utils.ts
ruocco-l: I'm dumb

oklinov (approved):

PR #33, opened 2025-10-22
Feat: skip building actors if only tests changed

### LINKED ISSUE #1: Skip building of actors if only tests changed

kazadoiyul:
Closes #1

### DIFF
--- a/bin/build.ts
+++ b/bin/build.ts
@@ -174,6 +174,13 @@ class ApifyBuilder {
             await this.apifyClient.build(build.id).delete();
         }
     }
+
+    async getDefaultBuilt() {
+        const client = this.apifyClient.actor(this.actorName)
+        const { defaultVersionNumber } = await this.getDefaultVersionAndTag()
+        const { id, actId, buildNumber } = await client.build(defaultVersionNumber)
+        return { buildId: id, actorId: actId, buildNumber, actorName: this.actorName };
+    }
 }
 
 type RunBuildsOptions = {
@@ -182,6 +189,15 @@ type RunBuildsOptions = {
     repoUrl: string
     branch: string
     dryRun: boolean
+}
+
+export const getAllDefaultBuilds = async (actorConfigs: ActorConfig[]) => {
+    const existingBuilds = await Promise.all(
+        actorConfigs.map(actor =>
+            ApifyBuilder.fromActorName(actor.actorName).getDefaultBuilt()
+        )
+    );
+    return existingBuilds;
 }
 
 export const runBuilds = async ({
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -10,7 +10,7 @@ import {
     spawnCommandInGhWorkspace,
     setCwd,
 } from './utils.js';
-import { runBuilds } from './build.js';
+import {getAllDefaultBuilds, runBuilds} from './build.js';
 import { getChangedFiles, getCommits } from './git.js';
 import { getPushData } from './github.js';
 import { notifyToSlack } from './slack.js';
@@ -107,7 +107,17 @@ await yargs()
                 branch: args.sourceBranch.replace('origin/', ''),
                 dryRun: args.dryRun,
             });
-            console.log(JSON.stringify(builds));
+
+            if (args.dryRun) {
+                console.log(JSON.stringify([...builds]));
+                return;
+            }
+
+            // add a build entry even if no build was triggered to run the tests on all actors
+            const remainingActors = actorConfigs.filter((a) => !actorsChanged.find((ac) => ac.actorName === a.actorName))
+            const existingBuilds = await getAllDefaultBuilds(remainingActors)
+
+            console.log(JSON.stringify([...builds, ...existingBuilds]));
         })
     .command(
         'release',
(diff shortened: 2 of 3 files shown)


bin/build.ts
```diff
@@ -174,6 +174,13 @@ class ApifyBuilder {
             await this.apifyClient.build(build.id).delete();
         }
     }
+
+    async getDefaultBuilt() {
```
oklinov, line 178: I think this isn't needed...you can extend `getDefaultVersionAndTag` to return `buildId`:
```ts
    const defaultBuildId = actorInfo.taggedBuilds![defaultBuildTag].buildId!;
```

bin/main.ts
```diff
@@ -10,7 +10,7 @@ import {
     spawnCommandInGhWorkspace,
     setCwd,
 } from './utils.js';
-import { runBuilds } from './build.js';
+import {getAllDefaultBuilds, runBuilds} from './build.js';
```
oklinov, line 13:
```suggestion
import { getAllDefaultBuilds, runBuilds } from './build.js';
```

bin/utils.ts
```diff
@@ -117,6 +117,11 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
     return IGNORED_TOP_LEVEL_FILES.some((ignoredFile) => sanitizedLowercaseFilePath.startsWith(ignoredFile));
 };
 
+const isIgnoredTestFile = (lowercaseFilePath: string) => {
```
oklinov, line 120: this should check if the `lowercaseFilePath` is a platform test, you can do it like this:
```ts
        return lowercaseFilePath.match(/^(code\/)?test\/platform\//);
```

I'd also rename it to `isPlatformTestFile`

bin/main.ts
```diff
@@ -107,7 +107,17 @@ await yargs()
...
                 dryRun: args.dryRun,
             });
-            console.log(JSON.stringify(builds));
+
+            if (args.dryRun) {
+                console.log(JSON.stringify([...builds]));
+                return;
+            }
+
+            // add a build entry even if no build was triggered to run the tests on all actors
```
oklinov, line 116: this will run always the tests for the actors that we're not building, no? My idea was to run them only if there is a changed in platform tests (in `test/platfrom`).

What I would do is to have `getChangedActors` return `actorsToTest` if there is such change. So, if we detect that there is a change in platform tests, we put all the actors that are not being built to `actorsToTest`.

I prepared some unit tests that hopefully explain the logic:
```ts
    test('only core tests changed', async () => {
        const FILES = [
            'test/platform/core/core.test.ts',
        ];

        const { actorsChanged, actorsToTest, codeChanged } = await getChangedActors({ actorConfigs: ACTOR_CONFIGS, isLatest: false, filepathsChanged: FILES });

        expect(actorsChanged).toEqual([]);
        expect(actorsToTest).toEqual(ACTOR_CONFIGS);
        expect(codeChanged).toBe(false);
    });

    test('core tests and src (excluding standalone actors) changed', async () => {
        const FILES = [
            'src/main.ts',
            'test/platform/core/core.test.ts',
        ];

        const { actorsChanged, actorsToTest, codeChanged } = await getChangedActors({ actorConfigs: ACTOR_CONFIGS, isLatest: false, filepathsChanged: FILES });

        expect(actorsChanged).toEqual(ACTOR_CONFIGS.filter((actor) => actor.isStandalone));
        expect(actorsToTest).toEqual(ACTOR_CONFIGS.filter((actor) => !actor.isStandalone));
        expect(codeChanged).toBe(false);
    });

    test('core tests and one input_schema changed', async () => {
        const FILES = [
            'test/platform/core/core.test.ts',
            'actors/lukaskrivka_testing-github-integration-1/.actor/INPUT_SCHEMA.json'
        ];

        const { actorsChanged, actorsToTest, codeChanged } = await getChangedActors({ actorConfigs: ACTOR_CONFIGS, isLatest: false, filepathsChanged: FILES });

        expect(actorsChanged).toEqual([ACTOR_CONFIGS[0]]);
        expect(actorsToTest).toEqual(ACTOR_CONFIGS.slice(1));
        expect(codeChanged).toBe(false);
    });
```

let me know if something's not clear

oklinov (changes requested):
sorry for the delay :slightly_frowning_face:

metalwarrior665:
This is a good feature to have but currently it is a bit hard to implement since we don't know what builds to use. Let's look at it after we finish the current round of refactors, most likely just to write it from scratch.

oklinov:
> ...since we don't know what builds to use

if the build step is skipped then we should use default (`latest`) builds, no?

metalwarrior665:
> if the build step is skipped then we should use default (latest) builds, no?

I can imagine you are in PR branch, do some changes and later update tests, you would want to run the tests vs last validated commit. So let's think this through. What was the use-case @kazadoiyul ?

oklinov:
this issue came from me... the use-case was: I created a PR with only changes in `/test/platform/` but it built everything.

> I can imagine you are in PR branch, do some changes and later update tests, you would want to run the tests vs last validated commit.

hmm, I didn't think about this case :thinking: , ideally we'd run the changed tests with the last built builds of that PR, but that complicates it a lot

PR #36, opened 2025-11-05
Feat: pretty slack messages

### LINKED ISSUE #19: Prettier slack messages
Image

ruocco-l:
Closes #19 
I've taken the liberty to do some changes on the `toFinishWith` matcher so that it tells something about what went wrong.

Some screenshots of how it looks:
Failed tests: 
Screenshot 2025-11-05 at 13 25 18
Markdown to be pushed on `notif-REPO`:
Screenshot 2025-11-05 at 13 29 24

Releases:
Logs:
Screenshot 2025-11-05 at 12 59 42
Markdown to be pushed on Slack:
On `delivery-public-actors`:
Screenshot 2025-11-05 at 12 59 53
On `notif-REPO`:
Screenshot 2025-11-05 at 13 00 03

---

Updated:
Now tests on slack will link to the run and use the name of the actor as anchor text
Screenshot 2025-11-24 at 14 28 06

### DIFF
--- a/bin/github.ts
+++ b/bin/github.ts
@@ -8,6 +8,7 @@ export const getPushData = async (path: string) => {
     const {
         commits: ghCommits,
         repository: { ssh_url: repoUrl, name },
+        head_commit: { author }
     } = event;
     const commits: Commit[] = ghCommits.map(({ author, message, id, timestamp }) => ({
         author: author.username,
@@ -24,6 +25,7 @@ export const getPushData = async (path: string) => {
         repoUrl,
         changelog,
         repository: name,
+        author: author.name
     };
 };
 
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -123,6 +123,7 @@ await yargs()
                 commits,
                 changelog,
                 repository,
+                author,
             } = await getPushData(args.pushEventPath);
             const isLatest = true;
             const actorConfigs = await getRepoActors();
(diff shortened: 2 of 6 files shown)


kazadoiyul:
i like the new logs!
just one thing: could you add the name of the actor for each failure? or maybe (again, if possible and not too complicated) write the name of the failed actor and then list all failures for it (then the same thing for all actors that have at least one failure)

i mean, for slack actor name can be next to each failure but for github it should be grouped (but it seems it is already grouped)

ruocco-l:
> could you add the name of the actor for each failure? 

I can easily add the name, even better I can make the name of the actor as the anchor text for the run link, so we save some space (on Slack)

> write the name of the failed actor and then list all failures for it 

I would rather not do this because, imo, tests should be grouped (if we want them to ) by test-type (core/logic/count...) not by actor. So, e.g., if you have a pagination problem you would have all you count tests fail, all grouped together, not the same pagination error "scattered" among the actors.

kazadoiyul:
oh, yeah, makes sense
i just thought about that as we mostly have test files split based on the actor but def makes sense to group it by type

oklinov (changes requested):
thanks. the report part looks good to me.

I'm not sure about the `toFinishWith` though...I'll have to check how it looks in vitest...could you please move it to a separate PR? we can discuss it there

oklinov (approved):

Issue #35, opened 2025-11-03
Do not auto-check maxRequestRetries <= 4 & allow overriding `toFinishWith`


foxt451:
Regarding `Do not auto-check maxRequestRetries <= 4 ` I think that it's quite restrictive, especially since on bigger runs with thousands of results, it's just impossible to have anything like that. Still, some repos might need it, and in such a case, it would be nice to somehow allow overriding default options of this checker (the only way I can think of now is to create some wrappers on your side, for which you always have to remember to use it then)

PR #38, opened 2025-11-19
feat: add maxRetriesPerRequest test

### LINKED ISSUE #37: Add maxFailingPerRequest
On TikTok we had the necessity to work this in
```ts
async function checkMaxRetryNumberOnSingleRequest(runResult: RunCommandOutput, MAX_RETRY_NUMBER: number = 10) {
    await expectAsync(runResult).withStatistics((stats) => {
        expect(stats.requestRetryHistogram.length)
            .withContext(runResult.format('Retry count on a single request'))
            .toBeLessThan(MAX_RETRY_NUMBER + 1);
    });
}
```
which basically makes sure that no request retry more than `MAX_RETRY_NUMBER`. We could work this in `toFinishWith`.

ruocco-l:
Closes #37 

This is one feature I would like to add. It is optional to check because I can't figure out a way to generalize a maximum (but this is something that can be overcome if we allow overriding defaults with #35 )

@kazadoiyul since this repo doesn't allow multiple reviewer (can we fix this?)

### DIFF
--- a/lib/extend-expect.ts
+++ b/lib/extend-expect.ts
@@ -196,6 +196,17 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
                 }
             }
 
+            if (options.maxRetriesPerRequest !== null) {
+                const { maxRetriesPerRequest } = options;
+                const requestRetryHistogram = stats?.requestRetryHistogram || [];
+                const maxRetriesObserved = requestRetryHistogram.length;
+                if (maxRetriesObserved > maxRetriesPerRequest) {
+                    diffs.pass = false;
+                    diffs.actual.push(`maxRetriesPerRequest=${maxRetriesObserved - 1}`);
+                    diffs.expected.push(`maxRetriesPerRequest=${maxRetriesPerRequest}`);
+                }
+            }
+
             return {
                 pass: diffs.pass,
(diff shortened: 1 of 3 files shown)


lib/extend-expect.ts
```diff
@@ -196,6 +196,17 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
                 }
             }
 
+            if (options.maxRetriesPerRequest !== null) {
+                const { maxRetriesPerRequest } = options;
+                const requestRetryHistogram = stats?.requestRetryHistogram || [];
+                const maxRetriesObserved = requestRetryHistogram.length;
```
oklinov: I think this isn't correct. Example:
- `stats.requestRetryHistogram = [ 177, 28 ]`, therefore `maxRetriesObserved = 2`
- `maxRetriesPerRequest` = `1`
- then it will fail, though it shoudn't: `maxRetriesObserved > maxRetriesPerRequest` will be true

You need to subtract 1 before the comparisson:
```suggestion
                const maxRetriesObserved = (stats?.requestRetryHistogram ?? [0]).length - 1;
```

oklinov (changes requested):
sorry for the late review

oklinov (approved):

PR #39, opened 2025-11-24
feat: verbose message for toFinishWith failed assertions

ruocco-l:
Reworks `toFinishWith` to push verbose explicative messages on what actually failed.

@kazadoiyul

### DIFF
--- a/lib/extend-expect.ts
+++ b/lib/extend-expect.ts
@@ -132,7 +132,7 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
                     diffs.pass = false;
                     diffs.actual.push(`status=${actual}`);
                     diffs.expected.push(`status=${expected}`);
-                    failedAssertions.push(`Failed status check.`);
+                    failedAssertions.push(`Failed status check, expected "${expected}", got "${actual}".`);
                 }
             }
 
--- a/lib/extend-expect.ts
+++ b/lib/extend-expect.ts
@@ -141,21 +141,23 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
                 stats: { durationMillis },
             } = await received.getRunInfo();
             const datasetItemCount = (await received.getDataset()).items.length;
-            const stats = (await received.getStatistics());
+            const stats = await received.getStatistics();
 
-            const datasetItemCountTest = isWithinInterval(diffs, datasetItemCount, options, 'datasetItemCount');
-            if (datasetItemCountTest === false) {
-                failedAssertions.push(`Failed dataset item count check.`);
-            }
-            const durationTest = isWithinInterval(diffs, durationMillis, options, 'duration');
-            if (durationTest === false) {
-                failedAssertions.push(`Failed duration check.`);
-            }
-            isWithinInterval(diffs, stats?.requestsFailed, options, 'failedRequests');
-            const requestsRetriesTest = isWithinInterval(diffs, stats?.requestsRetries, options, 'requestsRetries');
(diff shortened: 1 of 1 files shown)


lib/extend-expect.ts
```diff
@@ -193,12 +211,13 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
...
                     diffs.actual.push(` logs=[${occuredLogs.join(', ')}]`);
                     diffs.expected.push(` logs=[]`);
+                    failedAssertions.push(`Failed forbidden logs check.`);
                 }
             }
 
             return {
                 pass: diffs.pass,
-                message: () => `Run ${received.id} didn't finish as expected`,
+                message: () => `Run did not finish as expected. Failed assertions: ${failedAssertions.join(' ',)}`,
```
kazadoiyul, line 222: why did you remove runId from the log? is it present in failed assertions or is it just a bug?

kazadoiyul (approved):
i like more detailed log, thanks a lot!

lib/extend-expect.ts
ruocco-l: Yeah, that's because in the last PR we now get links to every run along with the failing actor name. Logs will look like this
```
AssertionError: Fake test to show input values in the report: expected false to be true // Object.is equality --- <https://console.apify.com/view/runs/pgP2uc6eKQgD6H1J7|streamers/youtube-channel-scraper>
        blocks: • Error: Run did not finish as expected. Failed assertions: Failed duration check. --- <https://console.apify.com/view/runs/pgP2uc6eKQgD6H1J7|streamers/youtube-channel-scraper>
                • AssertionError: Fake test to show input values in the report: expected false to be true // Object.is equality --- <https://console.apify.com/view/runs/Jwx49cEPs5fuoSg9d|streamers/youtube-comments-scraper>

```

foxt451 (commented):
Maybe not related to this PR, but could you also add logic to show the full diff from the underlying test lib? Like you can get `Error: Expected "undefined" to be a non-empty string` and have no idea what it means unless you navigate to the job

lib/extend-expect.ts
```diff
@@ -125,9 +127,13 @@ export const extendExpect = (expect: ExpectStatic): ExpectStatic => {
...
                 const actual = received.status;
-                diffs.pass = expected === actual;
-                diffs.actual.push(`status=${actual}`);
-                diffs.expected.push(`status=${expected}`);
+                const statusTest = expected === actual;
+                if (statusTest === false) {
+                    diffs.pass = false;
+                    diffs.actual.push(`status=${actual}`);
+                    diffs.expected.push(`status=${expected}`);
+                    failedAssertions.push(`Failed status check.`);
```
oklinov: optional:
```suggestion
                    failedAssertions.push(`Failed status check, expected "${expected}", got "${actual}".`);
```

we'd then get:
```
Run did not finish as expected. Failed assertions: Failed status check, expected "TIMING-OUT", got "SUCCEEDED".
```

oklinov (approved):
LGTM..I just left one optional suggestion...thanks and sorry for the late review

ruocco-l:
I added the `expected` vs `actual` in the failed assertion message. If there are multiple assertions failing (e.g. `datasetCount` and `ppeEventCount`, see screenshot) they are written in the same line and it's not the best, but adding other character or line breaks can be also confusing (and potentially tricky, since
 we are counting on the line break to create the status message). Maybe we can wait and see how they look on Slack when we have a run failing big.

Screenshot 2026-01-08 at 13 00 32

oklinov (approved):

PR #43, opened 2025-12-22
docs: update retry count default value to 1

Patai5:
Updates the JSDoc default value to 1, to reflect the actual value:
https://github.com/apify-projects/apify-test-tools/blob/d7e3a1ab2131a535d7bf4ece4e11cdf00544258b/lib/lib.ts#L56-L58

I got pretty confused by this 😭 A test was retrying for me because I didn't set `retry: 0` there, but the jsdoc was saying it should be 0, when in fact not 🙏 

---
Also, not sure about the reasoning behind the default value of 1 🤔 Personally I will set all my tests to 0, because tests should not be flaky, and if they are, you should know about it and fix it 🥀

### DIFF
--- a/lib/lib.ts
+++ b/lib/lib.ts
@@ -5,10 +5,9 @@ import {
     TestFunction,
     test as vitestTest,
     SuiteFactory,
-    TestOptions,
     TestContext,
 } from 'vitest';
-import type { ActorBuild, RunOptions } from './types';
+import type { ActorBuild, ActorTestOptions, RunOptions } from './types';
 import { RunTestResult } from './run-test-result.js';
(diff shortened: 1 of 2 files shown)


oklinov (approved):
lgtm, thanks

---

> Also, not sure about the reasoning behind the default value of 1 🤔 Personally I will set all my tests to 0, because tests should not be flaky, and if they are, you should know about it and fix it 🥀

in my experience, web scraping e2e tests are never 100% stable so we (me and @metalwarrior665 ) decided to do 1 retry by default.

metalwarrior665:
The testing Actor used to behave like this already. But I could see that we are in stable enough situation to remove this default.

PR #46, opened 2026-01-12
fix: change module resolution to nodenext and fix types

JuanGalilea:
[CONTEXT](https://apify.slack.com/archives/C033WM3PXLK/p1768215248102279)

Allow for projects that use this to have `nodenext` as `module` and `moduleResolution` and fix typing issues around expect and its extensions

The declaration file change was done according to [vitest docs](https://vitest.dev/guide/extending-matchers.html)

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+    "printWidth": 120,
+    "tabWidth": 4,
+    "singleQuote": true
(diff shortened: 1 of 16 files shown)


oklinov (approved):
thanks, lgtm...there are some conflicts with the master

PR #56, opened 2026-03-04
feat(bin): don't test on nonfunctional changes in json files

### LINKED ISSUE #54: Don't run tests on non-functional changes (e.g. decription)
This is as TODO in the specification but we never got to do it. Now there is a good case because after Claude pushing to master incident, we restricted non devs from bypassing and this unnecessarily blocks them because tests fail on description changes.

The specs say:
- Don't run tests on non functional changes (we need to parse the changes in input schema, actor.json, etc. and see if they are functional)
- If it is functional, we should still only run a prefilled core test for that miniactor so it is super fast (you can make separate issue for this)

https://apify.slack.com/archives/C059M9L8ZLP/p1770985909712419

metalwarrior665:
Was mostly vibecoded :) 

resolves https://github.com/apify-projects/apify-test-tools/issues/54


ruocco-l (approved):
LGTM

metalwarrior665:
Thanks for the review. I will wait for @oklinov since he will return shortly. 

I also want to do unify the naming of changes a bit, I'm thinking 3 types:
- functional (test)
- cosmetic (release only)
- ignored (just dev only stuff)

ruocco-l:
Feels much more cleaner after the last commit. Please just change the import in `should-built-and-test.test.ts` so that the unit test do not scream failure :)

bin/diff-changes.ts
```diff
@@ -0,0 +1,126 @@
...
+    filepathsChanged: string[];
+    actorConfigs: ActorConfig[];
+    isLatest?: boolean;
+    commits: Commit[];
+}
+
+export const maybeParseActorFolder = (lowercaseFilePath: string): { isActorFolder: true, actorName: string } | { isActorFolder: false } => {
+    const match = lowercaseFilePath.match(/^(?:standalone-)?actors\/([^/]+)\/.+/);
+    if (match) {
+        return { isActorFolder: true, actorName: match[1].replace('_', '/') };
```
oklinov: this will break on `google_maps_email_extractor_standby-contact-details-scraper` in https://github.com/apify-store/contact-info/tree/master/actors/google_maps_email_extractor_standby-contact-details-scraper ... we need to replace the last `_` in `actorName`

bin/git.ts
```diff
@@ -8,11 +8,13 @@ const GIT_LOG_FORMAT = ['%H', '%aN<%aE>', '%aD', '%s'].join(GIT_FORMAT_SEPARATOR
...
  */
 export const getChangedFiles = (commits: Commit[]) => {
-    const changedFiles = spawnCommandInGhWorkspace(
+    const changedFilesString = spawnCommandInGhWorkspace(
         `git diff --name-only ${commits[0].sha}~..${commits[commits.length - 1].sha}`,
     );
 
-    return changedFiles.split('\n');
+    const changedFiles = changedFilesString.split('\n');
+    console.error(`Changed files: ${changedFiles.join(', ')}`);
```
oklinov: do we need to log these? it'll polute the logs (probably not that much but still) 🤔

test/unit/parse-commit.test.ts
```diff
@@ -1,4 +1,5 @@
-import { describe, it, expect } from 'vitest';
+import { describe, expect,it } from 'vitest';
```
oklinov:
```suggestion
import { describe, expect, it } from 'vitest';
```

test/unit/should-built-and-test.test.ts
```diff
@@ -26,57 +26,51 @@ describe('Should build and test parser', () => {
     test('Ignores dev-only readme', async () => {
         const FILES = ['README.md', 'code/README.md', 'shared/README.md'];
 
-        const { actorsChanged, codeChanged } = await getChangedActors({ actorConfigs: ACTOR_CONFIGS, isLatest: false, filepathsChanged: FILES });
+        const { actorsChanged } = await getChangedActors({ actorConfigs: ACTOR_CONFIGS, isLatest: false, filepathsChanged: FILES });
```
oklinov: you forgot to update the import?

bin/diff-json-schema.ts
```diff
@@ -0,0 +1,43 @@
+import type { Commit } from './types.js';
+import { spawnCommandInGhWorkspace } from './utils.js';
+
+const COSMETIC_JSON_FIELD_NAMES = new Set([
+    'title', 'description', 'example', 'enumTitles', 'sectionCaption', 'sectionDescription',
```
oklinov, line 5:
```suggestion
    'title',
    'description',
    'example',
    'enumTitles',
    'sectionCaption',
    'sectionDescription',
```

oklinov (changes requested):

bin/diff-json-schema.ts
metalwarrior665: Will leave to future prettier

bin/git.ts
metalwarrior665: I was missing some logs in few debugging cases, I will show max 50. It doesn't really matter since no one should read the build step log unless for debugging

bin/diff-changes.ts
metalwarrior665: good catch ,that was broken for long time

test/unit/should-built-and-test.test.ts
metalwarrior665: Claude somehow forgot about this file, it needed a deeper refactor

metalwarrior665:
I refactored hopefully to simplify things. It seems to be working based on [this action](https://github.com/apify-store/testing-repo-for-github-actions/actions/runs/23499254354/job/68389329042?pr=42) but it is hard to test because base commit doesn't work.

Next steps:
- lint & stuff
- fix base commit & add more comments to the commits logic
- then new features

PR #60, opened 2026-03-17
feat: make slack notification explictly opt in

### LINKED ISSUE #55: Make slack report opt-in
As suggested [here](https://github.com/apify/store-website-content-crawler/pull/604#discussion_r2845912957) we can make the slack report opt-in.

Most repo are still using the actions in https://github.com/apify-store/github-actions-source/tree/new_master/.github/workflows so if we want to maintain the support for them without them noticing we can just update those workflow to include the appropriate opt-in flag

ruocco-l:
Closes #55 

Now `report-tests` and `release` commands must include `--notify-slack` or it will not trigger the notify notification, but just show the failed errors in the terminal.

### DIFF
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -144,7 +144,11 @@ await yargs()
             console.error(JSON.stringify(builds));
 
             // TODO: build circle actors
-            if (!notifySlack) return;
+            
+            if (!notifySlack) {
+                console.error(`Skipping slack notification. If you want to enable it, add --notify-slack flag and make sure SLACK_TOKEN_RELEASES_BOT env variable is set.`);
+            };
+
             await notifyToSlack({
                 changedFiles,
                 commits,
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -9,7 +9,7 @@ import { runBuilds } from './build.js';
 import { getChangedFiles, getCommits } from './git.js';
 import { getPushData } from './github.js';
 import { notifyToSlack } from './slack.js';
-import { reportTestRestuls } from './test-report.js';
+import { reportTestResults } from './test-report.js';
 
 /**
  * Middlewares to be run before every command execution
@@ -78,11 +78,12 @@ await yargs()
         (args) =>
             args
                 .option('report-file', { type: 'string', demandOption: true })
+                .option('notify-slack', { type: 'boolean', default: false })
                 .option('job-url', { type: 'string' })
                 .option('workflow-name', { type: 'string' })
                 .option('repository', { type: 'string' }),
         async (args) => {
-            await reportTestRestuls(args);
+            await reportTestResults(args);
         }
     )
     .command(
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -78,10 +78,9 @@ await yargs()
         (args) =>
             args
                 .option('report-file', { type: 'string', demandOption: true })
-                .option('notify-slack', { type: 'boolean', default: false })
+                .option('report-slack-channel', { type: 'string' })
                 .option('job-url', { type: 'string' })
-                .option('workflow-name', { type: 'string' })
-                .option('repository', { type: 'string' }),
+                .option('workflow-name', { type: 'string' }),
         async (args) => {
             await reportTestResults(args);
         }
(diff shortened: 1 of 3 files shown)


bin/main.ts
metalwarrior665: Why not just like this? (just format it after me :) ) Not sure why we would log error if we choose to not notify.
  
    
      
        Suggested change
      
    
    
      
          
            
            
          
          
            
                        
          
          
            
                        if (notifySlack) {
          
          
            
                            await notifyToSlack({
          
          
            
                                  changedFiles,
          
          
            
                                  commits,
          
          
            
                                  changelog,
          
          
            
                                  repository,
          
          
            
                                  dryRun,
          
          
            
                                  author,
          
          
            
                              });
          
          
            
                        };

bin/test-report.ts
```diff
@@ -78,16 +80,21 @@ export const reportTestRestuls = async ({
...
-
-    // remove repository-owner part
-    const channel = `#notif-${repository.replace(/[^/]+\//, '')}`;
-    console.error(`Sending a notification to slack channel ${channel}`);
+    const channel = repository ? `#notif-${repository.replace(/[^/]+\//, '')}` : null;
+    console.error(`Slack channel: ${channel ?? '(no repository provided)'}`);
     console.error('SLACK:', slackMessage);
     console.error('\tblocks:', blocks.join('\n\t\t'));
+
+    if (!notifySlack) {
```
metalwarrior665: I would move this above the channel selection as it is part of the slack logic

bin/main.ts
```diff
@@ -78,11 +78,12 @@ await yargs()
         (args) =>
             args
                 .option('report-file', { type: 'string', demandOption: true })
+                .option('notify-slack', { type: 'boolean', default: false })
```
metalwarrior665: If we would want to make this more usable outside of our team, we would probably need to make the Slack channels configurable etc. but for now this will do

metalwarrior665 (changes requested):
Looks good, just small nits

bin/main.ts
ruocco-l: I thought that since not sending the notification is the default, having a message that texplicitly says that you can actually send the notification can be informative, for either people that do not know how to set it up and people who simply forgot to add the flag

bin/test-report.ts
ruocco-l: Hm, I wanted to leave this in because I liked the message, but I guess the diff will be enough to evaluate the results 👍

bin/main.ts
oklinov: since this is already a breaking change, what if we added now 2 new cli arguments for defining slack channels instead of this flag?

So instead of

```sh
apify-test-tools --notify-slack
```

we'd have

```sh
apify-test-tools --report-slack-channel '#notif-google-maps' --release-slack-channel '#delivery-public-actor'
```

- `#delivery-public-actor` could be a variable at the org level
- `#notif-google-maps` we'd take it from the env: `"#notif-$(echo $GITHUB_REPOSITORY | cut -d '/' -f 2)"
  - use `cat` to remove the `apify-store/` part

In the GH workflow file, we would have:

```sh
npx apify-test-tools release --push-event-path ${{ github.event_path }} --release-slack-channel ${{ var.release_slack_channel }} --report-slack-channel "#notif-$(echo $GITHUB_REPOSITORY | cut -d "/" -f 2)"
```

metalwarrior665: I wouldn't worry about breaking changes, this is used only with us and we control https://github.com/apify-store/github-actions-source :) 

Feel free to go with the suggested change although I don't think it is really needed at this point. But it would be good to eventually look at the whole thing and try to restructure it so it is more generally usable. Like now it would still rely on the slack tokens in env.

ruocco-l: Personally I like passing the channel name instead of a `notify-slack` flag which feels very generic. That being said, I did the changes but I can just revert them if you think there is too much stuff going on for now :)

metalwarrior665 (approved):
Looks correct to me. We should figure out how to test this in the wild, probably manually adjust the beta version https://github.com/apify-store/testing-repo-for-github-actions and in the actions make sure we don't override the beta

metalwarrior665:
@ruocco-l Here is how to test it after you merge.

1. Since this starts with `feat`, it should make automatic beta npm release. see the [current ones](https://www.npmjs.com/package/apify-test-tools?activeTab=versions)
2. If it doesn't, let me know. You can release manually by going to Actions and selecting npm publish workflows on the left and fill master branch and beta target.
3. Then make a PR to https://github.com/apify-store/github-actions-source/ since these things need to go hand in hand.
4. Then you can test it manually with https://github.com/apify-store/testing-repo-for-github-actions. Change the source actions branch in workflows and change the apify-test-tools library to the new beta. I made it that the beta version is never overriden
5. Check it correctly uses the Slack channels. The testing repo has legit slack channel too


If you have idea how to automate this whole thing, that would be golden :D (AI might know)

PR #61, opened 2026-03-18
ci: Set up automated checks and npm publishing

janbuchar:
- As discussed with @ruocco-l, this adds automated code checks and npm publishing
- I'm pretty sure this won't work on the first try, but it's better to fix it when problems appear instead of wasting too much time trying to review it in depth

### DIFF
--- a/.github/scripts/before-beta-release.js
+++ b/.github/scripts/before-beta-release.js
@@ -1,0 +1,36 @@
+/* eslint-disable no-console */
+import { execSync } from 'node:child_process';
+import { readFileSync, writeFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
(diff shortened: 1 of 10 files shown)


metalwarrior665 (approved):
Thanks. I assume this is ± copied from our other packages.

Am I correct that it works as this?
- Every commit to master is beta package release
- Every commit with package.json version bump releases that version

janbuchar:
>     * Every commit to master is beta package release

Yes.

>     * Every commit with package.json version bump releases that version

No, package.json bumps are done on each commit to master, along with updating the changelog. Then, when you want to do a stable release, you manually dispatch the `manual_release_stable` workflow and that releases the version determined using conventional commits, or you can just specify the exact version yourself.

metalwarrior665:
Ok, so let's try this. The failing tests don't matter now, I will fix that later.

janbuchar:
The failing tests will probably prevent the publishing workflow from running though...

PR #70, opened 2026-03-26
fix(bin): in --base-commit, allow both sha string and Commit object

### LINKED ISSUE #67: Base commit logic doesn't work (and check commit logic thoroughly)
See here that base commit was available but was not picked up for the diff

https://github.com/apify-store/testing-repo-for-github-actions/actions/runs/23499254354/job/68389329042?pr=42

metalwarrior665:
Closes https://github.com/apify/apify-test-tools/issues/67 

Release new package version and after that merge this: https://github.com/apify-store/github-actions-source/pull/24

### DIFF
--- a/bin/git.ts
+++ b/bin/git.ts
@@ -17,11 +17,35 @@ export const getChangedFiles = (commits: Commit[]) => {
     return changedFiles;
 };
 
+const SHA_REGEX = /^[0-9a-f]{40}$/i;
+
+/**
+ *
+ * @param shaOrCommit Supports both a SHA string or a Commit object in JSON format. Can be empty.
+ * @returns The SHA string if valid, otherwise throws an error.
+ */
+export const parseBaseCommit = (shaOrCommit: string | undefined): string | undefined => {
+    if (!shaOrCommit) return undefined;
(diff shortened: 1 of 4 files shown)


bin/git.ts
```diff
@@ -17,11 +17,35 @@ export const getChangedFiles = (commits: Commit[]) => {
...
 
+const SHA_REGEX = /^[0-9a-f]{40}$/i;
+
+/**
+ *
+ * @param shaOrCommit Supports both a SHA string or a Commit object in JSON format. Can be empty.
+ * @returns The SHA string if valid, otherwise throws an error.
+ */
+export const parseBaseCommit = (shaOrCommit: string | undefined): string | undefined => {
+    if (!shaOrCommit) return undefined;
```
ruocco-l, line 28: Shouldn't we throw an error here?

ruocco-l (approved):
Looks ok, just a question on edgecase

bin/git.ts
metalwarrior665: I think we can keep it optional. Now the first commit in PR gets empty string

PR #78, opened 2026-04-08
perf(bin): don't test if we merge master into cosmetic only changes

metalwarrior665:
This is last optimization I had in backlog. Tested it on [this PR ](https://github.com/apify-store/testing-repo-for-github-actions/pull/44)

### DIFF
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -8,10 +8,11 @@ import { hideBin } from 'yargs/helpers';
 
 import { deleteOldBuilds, runBuilds } from './build.js';
 import { getChangedActors } from './diff-changes.js';
-import { getChangedFiles, getCommits } from './git.js';
+import { getBranchOnlyChangedFiles, getChangedFiles, getCommits, hasMergeFromTarget } from './git.js';
 import { getPushData } from './github.js';
 import { notifyToSlack } from './slack.js';
 import { reportTestResults } from './test-report.js';
+import type { Config } from './types.js';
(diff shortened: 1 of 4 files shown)


Patai5 (commented):
🥰

bin/main.ts
```diff
@@ -34,6 +35,44 @@ const buildOptions = (y: Argv) => {
...
+            '[MERGE-FROM-TARGET-OPTIMIZATION]: There is merge from target branch, checking if there are no functional changes in our own branch. If so, we can skip tests',
+        );
+        const branchOnlyFiles = getBranchOnlyChangedFiles(sourceBranch, targetBranch);
+        // Omit baseCommit to get full branch history. Validated functional commits can still interact with merged ones
+        const allBranchCommits = getCommits({ sourceBranch, targetBranch, baseCommit: undefined });
+        const branchOnlyActorsChanged = getChangedActors({
+            filepathsChanged: branchOnlyFiles,
+            actorConfigs,
+            commits: allBranchCommits,
+        });
```
ruocco-l, line 60: Claude spotted a "bug" where the tests are run if the branch has only cosmetic change in a `actors/*.json` file but the merge from target introduces a functional change on it. This will run the tests even though there is no non-validate interaction in the functional change. TBH I don't know how to fix it and it still feels like a small edge-case so I wouldn't bother with it.

ruocco-l (approved):

bin/main.ts
metalwarrior665: Good catch. But that will be very small number of these cases so I would not even bother with comment.

PR #83, opened 2026-05-11
feat: option to use docker cache on all builds

metalwarrior665:
I will turn it on in the source action after we merge.


ruocco-l (approved):
https://github.com/apify/apify-client-js/blob/94635b2cbe1e5aed76ba84a83b52de73e81d50bc/src/resource_clients/actor.ts#L268 I guess this JSDOC line is wrong then 🤔 (https://docs.apify.com/api/v2/act-builds-post API doc correctly say that it's false by default)

PR #96, opened 2026-06-24
feat!: replace folder naming convention for actor discovery with explicit config file

### LINKED ISSUE #84: Support .actor folder as source to build actor and test
As of now the library hinges on a very rigid folder structure: `actors/OWNER_ACTOR-NAME/`. or `standalone-actors/OWNER_ACTOR-NAME/`. This means that basic actor, like the ones from [templates](https://github.com/apify/actor-templates) cannot be tested with this tool.

Solving this will mean to find a reliable way to find the correct id (`owner/actor-name`) of the actor, knowing only the actor name from `actor.json`.

Possible ideas:
- Use a mandatory token like `OWNER_APIFY_TOKEN` that will also act as a fallback to the current folder structure (so if the APIFY_TOKEN_{OWNER} is not found it will fall back to `OWNER_APIFY_TOKEN`
- Require a config json (maybe configurable on repo variables?) that will match the name of the actors to their owner and the name of the token (still in the env variable). So we could have 
  ```typescript
  { actor-name1: {
      owner: "organization1",
      token: "BUILD_TOKEN"
  }}
  ```
     This solution will also relax the strict condition on the folder structure for the monorepos, allowing to quickly change ownership on Apify of an account, without actually change anything codewise (but it would be a breaking change)

ruocco-l:
Closes #84 

The CI tooling used to derive actor names and ownership entirely from folder naming conventions (`owner_actor-name`). This PR replaces that convention with an explicit, mandatory config file (`apify-test-tools.config.json`) at the repo root, and reworks change detection to scope by each actor's Docker build context instead of hardcoded `actors/`/`standalone-actors/` path regexes.

### The config file

Each repo declares its actors in `apify-test-tools.config.json`, one entry per actor: 
- `folder` (where the relevant .actor folder is), 
- `actorFullName` (`owner/name`, now the sole source of truth — no longer reconstructed from the folder),
- `tokenEnvVar` (no fallback, no derivation from owner/repo organization fails hard if unset),
- `overrideActorContext` for actors depending on paths outside their own folder. 

See the README for the full schema.

### Why change detection is scoped by Docker context instead of top-level directories

The old detection matched changed files against hardcoded `actors/`/`standalone-actors/` prefixes, which couldn't express "this actor also depends on `packages/shared`" or work for a single-actor repo with no `actors/` directory at all. Scoping by each actor's `dockerContextDir` (with `overrideActorContext` as an escape hatch) ties change detection to the same boundary Docker itself uses to build the actor, so the two can't drift apart. `.dockerignore` filtering follows the same reasoning: if Docker wouldn't see a file, it shouldn't trigger a rebuild either.

### Other design decisions

- **Actor identifier disambiguation** — `actorName` was used ambiguously for both the full `owner/name` string and the platform-internal ID, which produced hard-to-catch bugs when the wrong one was used. Split into `actorFullName`, `actorRawId`, and `actorId` throughout so each call site is unambiguous about which one it needs.
- **`ApifyBuilder.fromActorConfig`** replaces `fromActorName`, which re-derived the token independently from the actor name rather than trusting the already-resolved config.

### Breaking change

Repos must add `apify-test-tools.config.json` at the root, listing every actor with its `folder`, `actorFullName`, and `tokenEnvVar` (see README for the exact schema). The old folder-naming convention and `BUILDER_APIFY_TOKEN` fallback are both gone — there's no compatibility path for repos that don't add the config file. Special-case support for `CIRC_LE` accounts is also discontinued.


bin/diff-changes.ts
```diff
@@ -33,7 +31,8 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
         'eslint.config.mjs',
         '.prettierrc',
         '.editorconfig',
-        '.actor/',
+        // In root .actor/ mode, .actor/ changes must trigger builds
+        ...(isSingleActorRepo ? [] : ['.actor/']),
```
metalwarrior665: We can remove this completely, there is no reason we should commit changes to top level .actor in multiactor

bin/utils.ts
```diff
@@ -29,6 +31,63 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
     return value;
 };
 
+let cachedBuilderUsername: string | undefined;
+
+export const resolveBuilderTokenUsername = async (): Promise<string> => {
```
metalwarrior665: What is the use-case to need this? Just so you don't have to think about the username?

bin/utils.ts
```diff
@@ -29,6 +31,63 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+        );
+    }
+    const client = new ApifyClient({ token });
+    const user = await client.user().get();
+    cachedBuilderUsername = user.username!;
+    return cachedBuilderUsername;
+};
+
+const readActorName = async (actorJsonPath: string): Promise<string> => {
+    const actorJson: { name?: string } = JSON.parse(await fs.readFile(actorJsonPath, 'utf-8'));
```
metalwarrior665: I would rather introduce a new JSON file for our own custom needs than stitch non-spec fields to actor.json.

bin/utils.ts
```diff
@@ -48,17 +107,45 @@ export const getRepoActors = async (): Promise<ActorConfig[]> => {
...
+            throw new Error(
+                `Missing or unreadable .actor/actor.json in "${actorDir}". ` +
+                    `Every actor folder must contain .actor/actor.json with a "name" field.`,
+            );
         }
-        const [, folderType, owner, actorName] = match;
+
+        const folderName = actorDir.split('/')[1];
+        const folderType = actorDir.split('/')[0];
+        const owner = await resolveOwner(folderName);
```
metalwarrior665: Why not resolve owner from the same JSON place as the Actor name? I would just have `{ actorOwnerUsername, actorNameWithoutUsername }` or similar

bin/utils.ts
```diff
@@ -48,17 +107,45 @@ export const getRepoActors = async (): Promise<ActorConfig[]> => {
...
+        const fullName = `${owner}/${actorName}`;
+        const tokenEnvVar = resolveTokenEnvVar(owner);
+        await validateActorExists(fullName, tokenEnvVar);
+
         actorConfigs.push({
-            actorName: `${owner}/${actorName}`,
+            actorName: fullName,
             folder: actorDir,
             isStandalone: folderType === 'standalone-actors',
+            tokenEnvVar,
```
metalwarrior665: I'm not a big fan of passing the token around because it is more likely to leak, we should be able to just resolve it in place like we did no?

metalwarrior665 (changes requested):
I think this is good and obviously needed change. It won't be that big of a hassle for other teams since adding new Actors is quite rare. We need to polish this though.

metalwarrior665:
We should also sync with @gullmar about other repos. The monorepo where each Actor has its own src is also probably not ideally supported.

bin/utils.ts
```diff
@@ -48,17 +107,45 @@ export const getRepoActors = async (): Promise<ActorConfig[]> => {
         console.warn(`No /standalone-actors directory found in repo`);
         standaloneActorDirs = [];
     }
+    if (actorDirs.length === 0 && standaloneActorDirs.length === 0) {
+        let actorName: string;
+        try {
+            actorName = await readActorName('./.actor/actor.json');
+        } catch {
+            return [];
```
romanyaremchuk-apify: what do you think about adding some error handling here (logging or throw)? looks like it may fail silently

bin/build.ts
```diff
@@ -336,6 +324,7 @@ const findCircleApifyManaged = async (actorConfigs: ActorConfig[]): Promise<Acto
                     actorName: `${circleActor.username}/${circleActor.name}`,
                     folder: actorConfigFound.folder,
                     isStandalone: actorConfigFound.isStandalone,
+                    tokenEnvVar: 'APIFY_TOKEN_CIRC_LE',
```
romanyaremchuk-apify: should it be APIFY_TOKEN_CIRC_LE but not APIFY_TOKEN_CIRCLE? 🤔

romanyaremchuk-apify: btw 'APIFY_TOKEN_CIRC_LE' looks like it could be a const instead of a hardcoded string

bin/utils.ts
```diff
@@ -29,6 +31,63 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+        );
+    }
+    return actorJson.name;
+};
+
+const resolveOwner = async (folderName: string): Promise<string> => {
+    const ownerMatch = folderName.match(/^(.+)_[^_]+$/);
+    if (ownerMatch) return ownerMatch[1];
+    return resolveBuilderTokenUsername();
+};
```
Patai5: So the owner still has to be defined in the folder name 😅 That kind of defeats the purpose of this entire effort 🫠 

I would either want the conventionally named folder names only (owner + name in folder name), or to have both the owner and actor name in `actor.json`, but not both the conventionally named folder names for owners and `actor.json` for names at once 🙏

bin/utils.ts
```diff
@@ -29,6 +31,63 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+    return cachedBuilderUsername;
+};
+
+const readActorName = async (actorJsonPath: string): Promise<string> => {
+    const actorJson: { name?: string } = JSON.parse(await fs.readFile(actorJsonPath, 'utf-8'));
+    if (!actorJson.name) {
+        throw new Error(
+            `Missing "name" field in "${actorJsonPath}". ` +
+                `Every actor folder must have .actor/actor.json with a "name" field.`,
+        );
```
Patai5: This won't ever reach the console output, in both calls of this function it's just try catched and discarded.

I would rather just log a warning and return `null` out of the function. The caller can decide what to do in case it fails to get the name - that being `return [];` in call 1. and rethrowing the exact same error... in call 2.:
```ts
throw new Error(
	`Missing or unreadable .actor/actor.json in "${actorDir}". ` +
		`Every actor folder must contain .actor/actor.json with a "name" field.`,
);
```

bin/utils.ts
Patai5: +1

bin/utils.ts
Patai5: I'm personally against adding arbitrary fields to the `actor.json`. It's not a big deal, and we can add it there, I guess, but it feels dirty.

I would much rather be for the single source of truth config JSON file on the root level of the repo - e.g., `actors-configuration.json` and setup everything there. Alternatively, we could have this under each actor folder as `<MY_ACTOR_FOLDER>/actor-configuration.json`.

ruocco-l: The choice is basically a token fallback that deduces the owner or a config json (that explains where to build the actor and with which token). 

The reason I went with the former is because I don't like adding a config along with acotr.json and some other tokens. 

Since we have to do the token (TESTER and other token for the building), and the actor.json is already there, deducing the owner username allows not adding anything and in my opinion is easier to set up (because, again, it's a step less than adding a non-native property to the actor.json or creating a new file)

bin/utils.ts
ruocco-l: This doesn't pass the token, just the name of the env variable where it's stored

bin/utils.ts
ruocco-l: No. IF the owner is there we automatically assign and look for the appropriate token. If not we fallback to whatever is the owner of the BUILDER token. This is still useful if you have a default account where there are all the miniactors (for which is ok to fallback the BUILDER token) and have one miniactor that is build under a different account (for whatever reason), in which case you will specify it with the folder name `owner_whatever-actor-name-since-it's-not-used`.

But I agree that just adding the owner to the `actor.json` is much cleaner

bin/utils.ts
JuanGalilea: that would also enable us to cover miniactors, monorepos or single actor stuff without doing the 
`user-name_actor-name` convention.
It could all be config stuff

ruocco-l:
Alright alright, let's do the config

bin/utils.ts
Patai5: I agree that for most repos, the single `BUILDER` token would work fine, since we usually have everything under one account. So, for the most part, this would only have to cover edge cases. Even then it seems like there should be a better way to define the edge case then by folder name - be it `actor.json` or a global config file 🤔

metalwarrior665:
One more related long-term thing. When we designed this library, the core idea was "convention over configuration". We wanted everything to be at hardcoded places since that is always better than having to config and it worked well for that use-case.

Now if we want many (or arbitrary) different setups, we cannot keep patching it like this because we would have to keep adding `if`s to already complex logic. We have to rewrite most of the core logic to be truly configurable, e.g. each Actor has points to its "workspace" of path dirs (that can be shared).

On the other hand, the library has to remain somewhat opinionated, like what file changes trigger build etc, otherwise it has basically no value, it would just be some configs + few API calls.

I think for now, this PR is ok because top-level .actor is the template thing, but there probably already are things in the "changed Actors" that break. But for anything more, I would wait for th full rewrite. That should also make it truly open-source usable.

README.md
```diff
@@ -4,29 +4,102 @@
...
+npx apify-test-tools init-config --default-owner myteam --default-token-var APIFY_TOKEN_MYTEAM
+```
+
+The generated file looks like this:
 
+```json
+{
+    "actors": [
+        {
+            "folder": "actors/web-scraper",
```
metalwarrior665, line 24: I would add `actorName` too rather than deriving it from the folder

bin/main.ts
```diff
@@ -202,6 +202,23 @@ await yargs()
             });
         },
     )
+    .command(
```
metalwarrior665, line 206: I would remove this, Claude will one-shot this and you will want to manually check it anyway.

metalwarrior665 (changes requested):
I gave it another thought and I think it wouldn't be too hard to go all the way for really configurable config. Otherwise, we would have to potentially do another breaking change.

We should support (let's look of we can find more) at least 3 reference repos:
- typical Store monorepo
- template-like repo (e.g. WCC)
- monorepo with packages and actors importing them - [e-commerce](https://github.com/apify-professional-services/e-commerce-scraping-tool)

I'm thinking it could work like this:
1. Each Actor config points to the folder where .actor is (like now) or directly to the .actor folder.
2. Then from actor.json, we read `dockerContextDir` and that contains all folders/files that can influence the changes for that Actor. We can allow config to have `overrideActorContext` if they want to e.g. narrow it down.
3. Then instead of the current logic that if "code changes" we add mark all Actors as changed, we just do the changed files logic for each mini Actor independently, that cleans up the logic as well.

This way, we handle current Store monorepo (Actor folder + top-level context), template (just top-level), e-commerce (actor folder with its code, packages, shared)

What do you think?

README.md
ruocco-l: I thought about it, but I think can cause some confusion when you (usually when you publish the actor) play a little with the naming for SEO reasons (or similar). I think whatever is in the actor.json should be respected as the truth and not be overridden by some obscure logic from the testing package.

metalwarrior665: The testing lib will not set or change the name, it just check that it exists. The name.in actor.json is not a single source of truth so I would not use it, better to have all here

bin/utils.ts
```diff
@@ -29,53 +30,128 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+        config = JSON.parse(raw);
     } catch {
-        console.warn(`No /standalone-actors directory found in repo`);
-        standaloneActorDirs = [];
+        throw new Error(`Config file "${CONFIG_FILE_NAME}" contains invalid JSON.`);
     }
+
+    if (!Array.isArray(config.actors)) {
+        throw new Error(`Config file "${CONFIG_FILE_NAME}" must have an "actors" array at the top level.`);
+    }
```
Patai5, line 136: I would recommend using Zod for parsing the JSON config 👀

bin/main.ts
Patai5: I think this sort of CLI initialization utility perfectly fits into a library like here 🤔 

However unless we make this a completely seamless end to end migration command, then I'm also against it. Right now as far as I can tell, it doesn't really migrate the ENV vars. I will open up a separate comment for that.

bin/utils.ts
```diff
@@ -29,53 +30,128 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+
+    const entries: ActorConfigFileEntry[] = [];
+
+    for (const actorJsonPath of actorJsonFiles) {
+        const folder = actorJsonPath === '.actor/actor.json' ? '' : path.dirname(path.dirname(actorJsonPath));
+
+        entries.push({
+            folder: folder || '.',
+            owner: options.defaultOwner ?? '<OWNER>',
+            tokenEnvVar: options.defaultTokenVar ?? '<TOKEN_ENV_VAR>',
```
Patai5: We could make this be extracted from the current folder structure we use everywhere, so we would have a seamless end to end migration command 🤔 

E.g. take the owner username and translate it to the current env var name: `APIFY_TOKEN_<OWNER>`

bin/main.ts
Patai5: https://github.com/apify/apify-test-tools/pull/96/changes#r3482070978

Patai5 (approved):

bin/utils.ts
ruocco-l: I removed it.

bin/utils.ts
```diff
@@ -29,50 +30,102 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
-/**
- * Reads and parses all directories in `actors` directory
- * This works locally if checkoutRepoLocally is called first
- */
 export const getRepoActors = async (): Promise<ActorConfig[]> => {
-    let actorDirs: string[];
+    return readConfigFile();
+};
+
+const CONFIG_FILE_NAME = '.test-tools-actors-config.json';
```
oklinov: I'd call this `apify-test-tools.json` or `apify-test-tools.config.json` so it's clear it's for `apify-test-tools`

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
-          includes: 'all-actors' | ActorConfig;
-      };
-
-const classifyFileChange = (originalFilePath: string, actorConfigs: ActorConfig[], commits: Commit[]): FileChange => {
-    // Lowercase for case-insensitive matching; keep original for git show (case-sensitive on Linux)
+    | { impact: 'outside-context' }
+    | { impact: 'cosmetic'; semanticallyVerified: boolean }
+    | { impact: 'functional' };
+
+const isFileInContext = (lowercaseFilePath: string, actor: ActorConfig): boolean => {
```
metalwarrior665: Not sure why here it called it just actor :)

```suggestion
const isFileInContext = (lowercaseFilePath: string, actorConfig: ActorConfig): boolean => {
```

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
+    if (actor.overrideActorContext) {
```
metalwarrior665: I would resolve `overrideActorContext` at the config parsing so we carry over just the context path array, this function doesn't need to know if it is overriden or docker context

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
+): FileChangeForActor => {
     const lowercaseFilePath = originalFilePath.toLowerCase();
+
     if (isIgnoredTopLevelFile(lowercaseFilePath)) {
         return { impact: 'ignored' };
     }
 
-    if (lowercaseFilePath.endsWith('changelog.md')) {
-        return { impact: 'cosmetic', semanticallyVerified: false, includes: 'all-actors' };
+    if (!isFileInContext(lowercaseFilePath, actor)) {
```
metalwarrior665:
```suggestion
    if (!isFileInActorContext(lowercaseFilePath, actor)) {
```

bin/diff-changes.ts
```diff
@@ -8,22 +9,7 @@ interface ShouldBuildAndTestOptions {
...
-        // Some usernames weirdly use underscores, e.g. google_maps_email_extractor_standby-contact-details-scraper so we only need replace the last one
-        return { isActorFolder: true, actorName: match[1].replace(/_(?=[^_]*$)/, '/') };
-    }
-    return { isActorFolder: false };
-};
-
-/**
- * Also works for folders
- */
 const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
```
metalwarrior665: This function needs to be rewritten so the top level files are vs actor context so it works for standalone Actors too. Basically all the checks should be context aware. I think what we can simply do is to first "resolve context" by filtering the changed file paths from git and stripping the context path from them so they are "hoisted" to top level. E.g. if you have `standalone-actors/my-actor/readme.md` and context is `standalone-actors/my-actor` at the point we get changed file paths, we just strip this to just `readme.md` as it is relative to context for that Actor.

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
         '.prettierrc',
         '.editorconfig',
-        '.actor/',
     ];
-    // Strip out deprecated /code and /shared folders, treat them as top-level code
-    const sanitizedLowercaseFilePath = lowercaseFilePath.replace(/^code\//, '').replace(/^shared\//, '');
-
-    return IGNORED_TOP_LEVEL_FILES.some((ignoredFile) => sanitizedLowercaseFilePath.startsWith(ignoredFile));
+    // Strip deprecated code/ and shared/ prefixes — repos like apify-store/amazon use these
+    const sanitized = lowercaseFilePath.replace(/^code\//, '').replace(/^shared\//, '');
```
metalwarrior665: Follow-up to my previous comment, we can remove this hardcode and the repo owner will just have to add `code` and `shared` to context of all Actors

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
-    // Lowercase for case-insensitive matching; keep original for git show (case-sensitive on Linux)
+    | { impact: 'outside-context' }
+    | { impact: 'cosmetic'; semanticallyVerified: boolean }
+    | { impact: 'functional' };
+
+const isFileInContext = (lowercaseFilePath: string, actor: ActorConfig): boolean => {
+    if (actor.overrideActorContext) {
+        return actor.overrideActorContext.some((contextPath) => {
+            const lowerContextPath = contextPath.toLowerCase();
+            return lowerContextPath === '' || lowercaseFilePath.startsWith(`${lowerContextPath}/`);
```
metalwarrior665: Let's do some path parsing lib so we don't shoot ourselves in the foot with prefix/trailing slashes.

I would not require trailing slash, the context should be able to target files as well. They are included by default now if they are not in the ignore list but we might want to generalize that too in the future and I don't think there is any disadvantage

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
-                {
-                    actorName: actorFolderInfo.actorName,
-                    lowercaseFilePath,
-                },
-            );
-            return { impact: 'ignored' };
-        }
-        if (lowercaseFilePath.endsWith('readme.md')) {
-            return { impact: 'cosmetic', semanticallyVerified: false, includes: actorConfigChanged };
+    if (dockerIgnoreMatcher(originalFilePath)) {
```
metalwarrior665: Unrelated to the PR but we can add it, it is a small feature

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
+        return { impact: 'ignored' };
+    }
+
+    if (lowercaseFilePath.endsWith('readme.md') || lowercaseFilePath.endsWith('changelog.md')) {
+        return { impact: 'cosmetic', semanticallyVerified: false };
+    }
+
+    const lowerFolder = actor.folder.toLowerCase();
+    const isInActorFolder = lowerFolder === '' || lowercaseFilePath.startsWith(`${lowerFolder}/`);
```
metalwarrior665: We should verify by `.actor` path no? Let's require that all schemas and readme are inside .actor

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
+    if (lowercaseFilePath.endsWith('.json') && isInActorFolder) {
+        let isCosmetic = cosmeticCache.get(originalFilePath);
```
metalwarrior665: I don't see much value in the cache. I don't think there will be any real perf impact of those checks and also each Actor has their own schemas and their own context so we do pass over them only once anyway, right?

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
-    // For any other files, we assume they can interact with the code
-    return { impact: 'functional', includes: 'all-actors' };
+    return { impact: 'functional' };
+};
+
+/**
+ * Check if a file falls inside another actor's folder.
+ * Root actors (folder === "") never exclude files from siblings.
+ */
+const isExcludedBySibling = (lowercaseFilePath: string, actor: ActorConfig, allActors: ActorConfig[]): boolean => {
```
metalwarrior665, line 105: I think this can be simplified to just remove all .actor changes that aren't this Actor's .actor, we don't need to know what other Actors exist.

Theoretically, we could refactor to use the context to generate a list of file/directory paths for each Actor that would already exclude the sibling .actor and pass those resolved paths around instead of the context. This would allow us to do some filtering even before we know about changed files so the logic would be split into 2 stages, simplifying tests etc. We can prefilter ignored files, siblings, docker etc. Not a hard requirement but you can ask Claude what it thinks :)

bin/diff-changes.ts
```diff
@@ -33,64 +19,116 @@ const isIgnoredTopLevelFile = (lowercaseFilePath: string) => {
...
+
+const IMPACT_PRIORITY: Record<string, number> = { functional: 3, cosmetic: 2, ignored: 1 };
+
+/**
+ * A file can be classified differently by different actors (e.g. functional for a broad-context
+ * actor, outside-context for a narrow one). For logging we keep the most significant classification
+ * across all actors: functional > cosmetic > ignored. Files that are outside-context for every
+ * actor are treated as ignored.
+ */
+const updateFileImpact = (
```
metalwarrior665: I would change the logging, remove the impact priority (it's weird to override it like this) and forget the repo-wide approach and do it more in line with the Actor-centric approach
1. Group all changes that are same across Actors. Most changes will be shared by all.
2. Log changes for all groups separately. Something like "shared changes for Actors instagram-scraper, instagram-post-scraper: file1, file2", then "Changes specific to instagram-scraper: actors/instagram-scraper/.actor/input_schema.json" etc,

bin/utils.ts
```diff
@@ -29,50 +30,102 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+        try {
+            actorJson = JSON.parse(await fs.readFile(actorJsonPath, 'utf-8'));
+        } catch {
+            throw new Error(
+                `Cannot read "${actorJsonPath}". Every actor entry in "${CONFIG_FILE_NAME}" ` +
+                    `must have a corresponding .actor/actor.json file.`,
+            );
+        }
+
+        const actorDotDir = folder ? `${folder}/.actor` : '.actor';
```
metalwarrior665, line 190: I would just make folder required so it is explicit

metalwarrior665, line 190: We should validate that it has actor.json here, that will also catch some typos

bin/utils.ts
```diff
@@ -29,50 +30,102 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+            );
+        }
+
+        const normalizedDockerContextDir = dockerContextDir === '.' ? '' : dockerContextDir;
+
         actorConfigs.push({
-            actorName: `${owner}/${actorName}`,
-            folder: actorDir,
-            isStandalone: folderType === 'standalone-actors',
+            actorName: entry.actorName,
```
metalwarrior665:
```suggestion
            actorFullName: entry.actorName,
```

metalwarrior665 (changes requested):
I think we are going the right way, I have quite a few comments, the one with pre-filtering files has the biggest refactoring potential.

I don't think we are in any hurry so let's give this the time needed :)

bin/utils.ts
Patai5: +1

bin/diff-changes.ts
ruocco-l: ae0d1dc1dc7f577c7cb7d3e35ed5f8fbf332562e

bin/diff-changes.ts
ruocco-l: 18cf039 adresses the point, but it is then reworked in d29eb16

bin/diff-changes.ts
ruocco-l: faf4784, but, again, also d29eb167

bin/diff-changes.ts
ruocco-l: d29eb167abe06caa178114f9ba071c8bdc6a3dcc

bin/diff-changes.ts
ruocco-l: faf4784 and 9a3ec87. Readmes and changelog are usually sibling of .actor

bin/diff-changes.ts
ruocco-l: b3df6fa

bin/diff-changes.ts
ruocco-l: I didn't do it, it feels like a small improvement and I personally like the flow, it's simple enough

bin/diff-changes.ts
ruocco-l: 6980b49d02bd201b16fdc1cfe7035db26e57374e much better imo, thank you for the input

bin/utils.ts
ruocco-l: 73a2d8d1d95bedd9782eef46c9fc7893459c1fe3

ruocco-l: This is basically already there, the code tries to parse the json and if it can't it throws

bin/utils.ts
ruocco-l: This is actually something that I have been struggling since the start of this PR. The problem with this is that there is some incosistency with what is `actorName` and what is `actorId` (in both Apify and this project). I took the liberty of doing one single commit where I try to unify the names so we now have `actorFullName` for the human-readable name, `actorRawId` for the platform's ID, and `actorId` as the Apify conventional id (both raw id and full name ).

bin/utils.ts
ruocco-l: cbed5e2

bin/utils.ts
ruocco-l: I understand that this can be a confusing change, so maybe we can export it in a seprate pr, or just revert it and leave it as is

bin/utils.ts
```diff
@@ -29,50 +31,139 @@ export const getEnvVar = (varName: string, defaultValue?: string): string => {
...
+
+        const overlap = findOverlappingContextPaths(contextPaths);
+        if (overlap) {
+            throw new Error(
+                `Invalid context paths for folder "${entry.folder}" in "${CONFIG_FILE_NAME}": ` +
+                    `"${overlap[0]}" and "${overlap[1]}" overlap. Context paths must not be prefixes of one another.`,
+            );
+        }
+
+        if (!contextPaths.some((contextPath) => isPathWithinScope(folder, contextPath))) {
```
metalwarrior665: Do we need this? We carry over actorConfig.folder so we always know where our .actor lives. So we shouldn't need it inside the context paths

bin/diff-changes.ts
```diff
@@ -8,89 +10,145 @@ interface ShouldBuildAndTestOptions {
...
+    }
+
+    const hoistedFilePath = hoistPath(lowercaseFilePath, matchedContext);
+    if (isIgnoredTopLevelFile(hoistedFilePath)) {
         return { impact: 'ignored' };
     }
 
-    if (lowercaseFilePath.endsWith('changelog.md')) {
-        return { impact: 'cosmetic', semanticallyVerified: false, includes: 'all-actors' };
+    if (dockerIgnoreMatcher(originalFilePath)) {
```
metalwarrior665: .actor can be legit inside .dockerignore as it is evaluated by Apify platform before Docker build. That can be used to make Dockerfile more cacheable if you don't want to COPY individual folders like we d now. So either we skip it from dockerIgnore or we handle it before/special path.

bin/diff-changes.ts
```diff
@@ -8,89 +10,145 @@ interface ShouldBuildAndTestOptions {
...
-                {
-                    actorName: actorFolderInfo.actorName,
-                    lowercaseFilePath,
-                },
-            );
-            return { impact: 'ignored' };
-        }
-        if (lowercaseFilePath.endsWith('readme.md')) {
-            return { impact: 'cosmetic', semanticallyVerified: false, includes: actorConfigChanged };
+    const lowerFolder = actorConfig.folder.toLowerCase();
```
metalwarrior665:
```suggestion
    const lowerCaseFolder = actorConfig.folder.toLowerCase();
```

bin/diff-changes.ts
```diff
@@ -8,89 +10,145 @@ interface ShouldBuildAndTestOptions {
...
 
 /**
- * Also works for folders
+ * Classify a single file change for a single actor.
+ *
+ * Steps (in order):
+ * 1. Context matching (actorConfig.contextPaths) → outside-context if no match
+ * 2. Hardcoded ignore list, checked against the path hoisted relative to the matched context entry → ignored
+ * 3. .dockerignore filtering (patterns relative to dockerContextDir) → ignored if matched
+ * 4. README/CHANGELOG by filename → cosmetic if inside the actor's own folder, otherwise ignored
```
metalwarrior665: The Changelog change is wrong, in Store Actors, there is single shared changelog so it works differently from readme.

Instead of speculating what's the right setup, let's also do this more properly for all the Actor-only files. File is specific Actor related if:
- Is inside Actor folder (inside .actor or next to it) - We already handle this
- Is referenced from actor.json - if the file is outside of Actor folder but is referenced from actor.json, then it belongs to this Actor (and independenctly can belong to other Actor). So top-level Changelog.md will independently trigger change for every Actor (that points to it). We should parse this at the config file read. I think we can have a hardcoded list of files supported this way (readme, changelog, all schemas)

The ignored/cosmetic/functional logic depending on being test vs release and JSON file changes stays

EDIT: I'm dragging you through this PR a bit so feel free to just hotfix this with the old semantics and make TODO + issue for next iteration

bin/diff-changes.ts
```diff
@@ -8,89 +10,145 @@ interface ShouldBuildAndTestOptions {
...
-            );
-            return { impact: 'ignored' };
-        }
-        if (lowercaseFilePath.endsWith('readme.md')) {
-            return { impact: 'cosmetic', semanticallyVerified: false, includes: actorConfigChanged };
+    const lowerFolder = actorConfig.folder.toLowerCase();
+    const isInActorFolder = isPathWithinScope(lowercaseFilePath, lowerFolder);
+
+    if (lowercaseFilePath.endsWith('readme.md') || lowercaseFilePath.endsWith('changelog.md')) {
+        return isInActorFolder ? { impact: 'cosmetic', semanticallyVerified: false } : { impact: 'ignored' };
```
metalwarrior665, line 88: This needs fixing as mentioned [above](https://github.com/apify/apify-test-tools/pull/96/changes#r3552246484)

metalwarrior665 (changes requested):
Thanks for another rewrite! I think we are almost done.

metalwarrior665:
Also pls update the PR title :)

bin/dockerignore.ts
```diff
@@ -0,0 +1,39 @@
...
+export type DockerIgnoreMatcher = (repoRelativePath: string) => boolean;
+
+/**
+ * Load .dockerignore from the root of an actor's dockerContextDir and return a matcher
+ * that accepts repo-root-relative file paths. Patterns are resolved relative to
+ * dockerContextDir, matching Docker's own behavior.
+ *
+ * Returns a no-op matcher (always returns false) when the file is absent.
+ */
+export const loadDockerIgnore = (dockerContextDir: string): DockerIgnoreMatcher => {
```
gytelio: I found this bug. loadDockerIgnore doesn't normalize a leading ./ (e.g. ./node_modules) before passing patterns to the ignore package, so any .dockerignore written in that common style (like ./node_modules, ./test) silently matches nothing :/

test/unit/bin/dockerignore.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        readFileSync: vi.fn(),
+    },
+}));
+
+vi.mock('node:fs', () => ({ default: fsMock }));
+
+afterEach(() => vi.restoreAllMocks());
+
+describe('loadDockerIgnore', () => {
+    it('returns no-op matcher when .dockerignore is absent', () => {
```
gytelio, line 18: lets add a unit test for ./-prefixed pattern

gytelio: also edge cases like trailing slash

bin/dockerignore.ts
ruocco-l: Addressed here https://github.com/apify/apify-test-tools/pull/96/commits/48b7847455a2acf4f935c26b200320270c935a29

bin/diff-changes.ts
ruocco-l: https://github.com/apify/apify-test-tools/pull/96/changes/98c610bcf404b30b206029ce691098aa1e92bb42

bin/diff-changes.ts
ruocco-l: https://github.com/apify/apify-test-tools/pull/96/changes/48b7847455a2acf4f935c26b200320270c935a29

bin/utils.ts
ruocco-l: I think of the context as "I need to track the files in these paths because that's what this actor care about". With this logic it looks correct to me that the .actor folder should always be listed/in the scope.

The middle ground for this would be to always forcefully add the `folder` path to the overridden context list, what do you think?

JuanGalilea:
>Then from actor.json, we read dockerContextDir and that contains all folders/files that can influence the changes for that Actor. We can allow config to have overrideActorContext if they want to e.g. narrow it down.

@metalwarrior665 what would be the usecase of `overrideActorContext`?
Shouldn't `actor.json` have it right? what does `narrowing` look like?
I cant imagine a usecase where we would want to build with root as a different directory than the one specified by `actor.json`

ruocco-l:
> Shouldn't `actor.json` have it right? what does `narrowing` look like? I cant imagine a usecase where we would want to build with root as a different directory than the one specified by `actor.json`

`actor.json` can have a docker context that for multiple reasons could be too broad (e.g. you have only one dockerfile that works for all the miniactors) but for some miniactors may actually need only a subset of shared files.
```
actors/
├─ actor_1/
├─ actor_2/
├─ actor_3/
Dockerfile
shared/
├─ shared_1-2/
├─ shared_1-3/
├─ shared_1-2-3/
```
With this folder structure you will have to have the dockercontext to look at the whole `shared` folder, but with `overrideActorContext `you tell change-detection which subset of shared/ actually matters to actor_1 (even though Docker will still build the whole thing)

metalwarrior665:
> what would be the usecase of overrideActorContext?
Shouldn't actor.json have it right? what does narrowing look like?
I cant imagine a usecase where we would want to build with root as a different directory than the one specified by actor.json

This is not about building but about deciding if a change in a file should rebuilt & retest a specific Actor. In e-commerce monorepo, most folders are totally unrelated even if they are in build context

bin/utils.ts
metalwarrior665: I would add it automatically. I take as a special type of context that is always included but if you need it add explicitly, then for sure

metalwarrior665 (approved):
Great, I consider it done! Thanks for going through all.

Now the release will be a bit tricky. I published a beta apify-test-tools 0.9.0-beta.0. We cannot merge this because we automatically use latest and would break everyone. So until all repos migrate to this, we will have to keep rebasing this on top of master. But the migration should be simple.

Before starting the Store-wide migration, let's wait a few days if @Patai5 @JuanGalilea @oklinov will have some remarks.

I would not release this as 1.0. I would reserve that for the final public API that will be easy to use but any publisher and we will not break that for long time to come.

romanyaremchuk-apify (approved):

bin/utils.ts
ruocco-l: I guess you forgot to remove these comments from a month ago before approving :D Thank you for your review!

PR #100, opened 2026-07-07
feat: build zip from local source

### LINKED ISSUE #93: Better support for running platform tests locally
Currently, you can run platform tests locally but it is a bit cumbersome. You have to push to a branch (potentially new branch if you don't want to pollute the one you are working on), then use that branch to build Actors and pass those builds to tests. It is doable with a simple script but it is still harder than it needs to be.

We can improve it with 2 solutions (split into separate issues if you want):

1. Deploy code from src as zip file. We can either just build it into the standard 0.99 version or create a separate 0.98 just for this, so we don't confuse the git/zip sources. This should be quite simple change.
2. Run the actual Actors locally to with `apify run`. This will be a bit more involved and won't have the platform guarantees but might be helpful for very short checks

### LINKED ISSUE #94: Allow filtering or hardcoding Actors to build (useful locally)
If you want to initiate build locally, you have to rely on the "changed Actors" magic that relies on git. That often builds all Actors. But often you might need to quickly iterate only on a single miniactor and build just that.

Few options:
1. New build-actor command
2. Add a filter to the current build (would still respect changed Actors logic)

gytelio:
Closes #93 
Closes #94 

Adds a build-from-local CLI command that builds an Actor straight from local source files, no git push needed. Files are selected using git ls-files/check-ignore (so .gitignore rules apply automatically), and known secret-file patterns (.env, .pem, .key, etc.) are always dropped. Monorepo actors get their Docker context flattened into a temp dir first. Remaining files are read as text/base64 and uploaded as a SOURCE_FILES actor version, then built on the platform.

Tested locally using npm link

### DIFF
--- a/bin/build.ts
+++ b/bin/build.ts
@@ -10,7 +10,18 @@ import { ACTOR_SOURCE_TYPES } from '@apify/consts';
 import type { ActorConfig, BuildData } from './types.js';
 import { collectFilePaths, getGitignoredPaths, isOutsideDir, toSourceFile } from './utils.js';
 
-const SKIP_DIRS = new Set(['node_modules', '.git', 'apify_storage', 'dist', 'build', 'out', '.next', '.cache']);
+const SKIP_DIRS = new Set([
+    'node_modules',
+    '.git',
+    'apify_storage',
+    'storage',
+    'crawlee_storage',
+    'dist',
+    'build',
+    'out',
+    '.next',
+    '.cache',
+]);
 
 // JUST IN CASE. File patterns that commonly hold credentials — never ship these into a build, regardless
 // of sourceType or of whether the repo's .gitignore happens to list them. Everything else that should be
@@ -18,8 +29,6 @@ const SKIP_DIRS = new Set(['node_modules', '.git', 'apify_storage', 'dist', 'bui
 // in the repo's .gitignore — see getGitignoredPaths.
 const SKIP_FILE_PATTERNS = [/^\.env(\..+)?$/, /\.pem$/, /\.key$/, /\.pfx$/, /\.p12$/];
 const isSecretFile = (fileName: string): boolean => SKIP_FILE_PATTERNS.some((pattern) => pattern.test(fileName));
-
-const MAX_SOURCE_FILES_BYTES = 3 * 1024 * 1024;
 
 type BuildPrActorOptions = {
     buildTag?: string;
@@ -178,7 +187,6 @@ export class ApifyBuilder {
                 filePaths.map(async (filePath) => toSourceFile(filePath, sourceRootDir)),
             );
 
-            this.assertWithinSizeLimit(sourceFiles);
             return sourceFiles;
         } finally {
             // Only the flattened copy is temporary — never delete the actor's own directory.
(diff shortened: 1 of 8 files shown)


bin/build.ts
```diff
@@ -1,9 +1,34 @@
...
+
+import type { ActorVersionSourceFile, Build } from 'apify-client';
 import { ApifyClient } from 'apify-client';
 
 import { ACTOR_SOURCE_TYPES } from '@apify/consts';
 
 import type { ActorConfig, BuildData } from './types.js';
+import { collectFilePaths, getGitignoredPaths, isOutsideDir, toSourceFile } from './utils.js';
+
+const SKIP_DIRS = new Set([
```
metalwarrior665: I think we could just use .gitignore?

bin/build.ts
```diff
@@ -101,15 +126,205 @@ class ApifyBuilder {
...
+        } else {
+            await actorClient.version(ZIP_VERSION).update(actorVersion);
+        }
+
+        const { id, actId, buildNumber } = await actorClient.build(ZIP_VERSION, { useCache: false });
+        console.error(`[${this.actorName}]: ${id} (${buildNumber})`);
+        return { buildId: id, actorId: actId, buildNumber, actorName: this.actorName };
+    };
+
+    collectSourceFiles = async (actorDir: string): Promise<ActorVersionSourceFile[]> => {
```
metalwarrior665: Can you move all these Zip push related helpers to a new file? It is too much and production path doesn't use it so let's not pollute this file

bin/utils.ts
```diff
@@ -1,8 +1,65 @@
...
+import { SOURCE_FILE_FORMATS } from '@apify/consts';
 
 import type { ActorConfig } from './types.js';
 
+// Returns true when `childPath` is not inside `parentPath`.
+// Used to detect monorepo actors whose dockerContextDir escapes the actor directory.
+export const isOutsideDir = (childPath: string, parentPath: string): boolean =>
+    path.relative(parentPath, childPath).startsWith('..');
+
+export const collectFilePaths = async (rootDir: string, skipDirs: Set<string>): Promise<string[]> => {
```
metalwarrior665: Can we call git to get tracked (+ untracked in working fir) files instead of making our own func?

bin/utils.ts
```diff
@@ -1,8 +1,65 @@
...
+    if (result.status !== 0 && result.status !== 1) {
+        throw new Error(`[Command failed]: git check-ignore\n${result.stderr.toString()}`);
+    }
+
+    return new Set(result.stdout.toString().split('\n').filter(Boolean));
+};
+
+const isBinary = (buffer: Buffer): boolean => buffer.includes(0);
+
+export const toSourceFile = async (absPath: string, rootDir: string): Promise<ActorVersionSourceFile> => {
```
metalwarrior665: Or move these to separate zip utils so it is clearer what it is supporting

```suggestion
export const toActorVersionSourceFile = async (absPath: string, rootDir: string): Promise<ActorVersionSourceFile> => {
```

bin/main.ts
```diff
@@ -202,6 +202,31 @@ await yargs()
             });
         },
     )
+    .command(
+        'build-zip',
```
metalwarrior665: Maybe clearer?

```suggestion
        'build-from-local',
```

bin/utils.ts
metalwarrior665: Also maybe check what CLI does on `apify push`

bin/build.ts
```diff
@@ -101,15 +126,205 @@ class ApifyBuilder {
...
+    // collect the actor directory of a monorepo actor — the platform would reject any path
+    // escaping it. Fix: create a temporary "flattened" directory where:
+    //   - the Docker context's non-ignored files (repo root) are copied to the temp dir root
+    //   - the actor's .actor/ directory is overlaid at the temp dir root (through the same
+    //     gitignore/secret-pattern filter as the rest of the context — see collectNonIgnoredFiles)
+    //   - actor.json path fields are rewritten to be relative to the new location
+    //
+    // Result: the collected root IS the Docker context, .actor/ is at that root, and
+    // all relative paths (dockerfile, dockerContextDir, changelog) are exactly one
+    // level up ("..") instead of three ("../../..").
```
metalwarrior665: Uh, that's a lot of workarounding here but i guess no other way now with source files

We also need to keep the original .actor because some files will read from it, e.g. `from '../actors/apify_instagram-scraper/.actor/input_schema.json'`

metalwarrior665 (changes requested):
I think it is pretty close to done (if it works :) ) Just minor remarks

bin/utils.ts
gytelio: Yeah good idea with the git tracked. I implemented it. For apify cli I checked we are doing here pretty much the same, the only thing is not supported is actorignore.

bin/build.ts
gytelio: Yeah, flattenMonorepoContext always keeps the og .actor

metalwarrior665 (approved):
Nice. There shouldn't be any potential breakage of existing code so if you tested this locally, I will just release it.

PR #104, opened 2026-07-15
fix: dockerignore

gytelio:
We need to ignore dockerignore files same way we ignore gitignore. We need to ignore directories also like /tests, because otherwise we are getting "sourceFilesTooLarge\ error. 
Found in `@apify/consts ` :
`MAX_MULTIFILE_BYTES = 3 * 1024 ** 2   // 3 MiB (3,145,728 bytes)`

### DIFF
--- a/bin/build-from-local.ts
+++ b/bin/build-from-local.ts
@@ -6,7 +6,13 @@ import type { ActorVersionSourceFile } from 'apify-client';
 
 import { ApifyBuilder, waitAndSummarizeBuilds } from './build.js';
 import type { ActorConfig, BuildData } from './types.js';
-import { getGitignoredPaths, isOutsideDir, listRepoFilePaths, toActorVersionSourceFile } from './utils.js';
+import {
+    getDockerignoredPaths,
+    getGitignoredPaths,
+    isOutsideDir,
+    listRepoFilePaths,
+    toActorVersionSourceFile,
+} from './utils.js';
 
 // JUST IN CASE. File patterns that commonly hold credentials — never ship these into a build, regardless
 // of sourceType or of whether the repo's .gitignore happens to list them. Everything else that should be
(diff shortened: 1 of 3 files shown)


metalwarrior665 (changes requested):
We are already implementing dockerignore handling in https://github.com/apify/apify-test-tools/pull/96

These are for different features so that would be fine but we would have duplicate utils for dockerignore. Can you check how it is handled in that PR and if you could reuse it or if it does overlap?

gytelio:
> We are already implementing dockerignore handling in #96
> 
> These are for different features so that would be fine but we would have duplicate utils for dockerignore. Can you check how it is handled in that PR and if you could reuse it or if it does overlap?

@metalwarrior665 I checked. The util function from #96 I cannot just plug it, the function and this one expect different kinds of input. But I would like to reuse it, I like the usage of the ignore package. But there is a bug that I was trying to fix with this one  (I added a comment) https://github.com/apify/apify-test-tools/pull/96/changes#r3594171691. 

I would like to do a cleanup after the 96 will be merged and the bug is fixed.  But currently I would like to merge this one because it works and its tested, and some people currently wants to use the local build now.

ruocco-l:
@gytelio thank you for your comment on my PR. I think, as @metalwarrior665 already pointed out in my PR, this is a feature that astray from my original intent. I think we should go ahead and implement this PR first and then it would be my work to make it work in my PR, instead of doing it the other way around.

metalwarrior665:
Let's give that PR few days to get merged. Merging this one first would complicate it. But if it gets stuck then let's do it.

metalwarrior665:
Ok, so if @ruocco-l is ok rebasing on top of this one, we can go with it

metalwarrior665 (approved):

gytelio:
@metalwarrior665 Im not familiar with this. If I merge it to main will it automatically be on the latest release? Or do I need to run something manually?

metalwarrior665:
It will release a beta version. For regular version, you need Actions -> Stable release (patch version). I can do it

metalwarrior665:
Released as 0.8.6

PR #108, opened 2026-08-04
feat(only-and-omit): pick or exclude actors

JuanGalilea:
added `--omit-actors` and `--only-actors` to ignore or focus on certain actors.

Purpose is to enable the 2 step deployment process of POD E.
pod e needs:
1. deploy everything but the main actor
2. deploy the main actor

Even if not needed for this case, added both as arrays. So it can be used as:
`--omit-actors actor1 actor2 actor3 ...`

Also the filtering function asserts all being found on the config. Since there is probably something wrong if your options don't match your config.

This is the most versatile thing i cooked up and then the responsibility of the ordering and such falls on the deployment action workflow. Which in this case would be:
1. deploy with omit main actor
2. deploy only main actor

### DIFF
--- a/bin/main.ts
+++ b/bin/main.ts
@@ -6,7 +6,7 @@ import yargs, { type Argv } from 'yargs';
 // eslint-disable-next-line import/extensions --- With .js, it cannot find types
 import { hideBin } from 'yargs/helpers';
 
-import { filterActorByConfig } from './actor-filtering.js';
+import { selectActors } from './actor-filtering.js';
 import { deleteOldBuilds, runBuilds } from './build.js';
 import { runBuildsFromLocal } from './build-from-local.js';
 import { getChangedActors } from './diff-changes.js';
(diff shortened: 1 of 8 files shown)


bin/main.ts
```diff
@@ -37,27 +37,44 @@ const buildOptions = (y: Argv) => {
...
-) => {
-    const actorConfigs = await readConfigFile();
+/**
+ * Actor-selection flags, applied to every command that reads the actor config so a caller can
+ * narrow the set it operates on (e.g. two-stage releases: `--ignore X`, then `--actors X`).
+ * Kept separate from `buildOptions` so the read-only git commands don't advertise flags they ignore.
+ */
+export const actorSelectionOptions = <T>(y: Argv<T>) => {
+    return y
+        .option('actors', {
```
ruocco-l, line 51: NIT: I'd rather call these `--include-actors` and `--omit-actors`

ruocco-l (approved):
I'm fine with it. The fact that e-commerce already have this problem and just adding this fix it is a sign that can be useful in multiple workflows (the build from local selection is a similar example). Thanks!

bin/main.ts
JuanGalilea: only went by these since local build already used `--actors`.
Can change it if needed tho, i don't really care either way

metalwarrior665 (approved):
Looks generally useful (since we want to open this more to public) and low maintenance overhead. Ideally rebase it later so that this is separate commit to master after the breaking one from Luigi.

PR #120, opened 2026-08-20
feat(workflows): host the reusable GitHub workflows in this repo

metalwarrior665:
This PR introduces a complete set of reusable GitHub workflows and supporting infrastructure that consumer repositories can use for testing, building, and releasing their Actors.

## Summary

The package now includes production-ready reusable workflows alongside the CLI, enabling consumer repos to standardize their CI/CD pipelines. These workflows handle platform testing, PR validation, releases, and Claude-powered automated investigation and fixes for failing tests.

## Key Changes

- **Reusable Workflows** (`.github/workflows/`):
  - `pr-build-test.yaml`: Runs unit tests and platform tests on pull requests
  - `platform-tests.yaml`: Scheduled platform test runs with Slack reporting
  - `push-build-latest.yaml`: Builds and releases Actors on master push
  - `claude.yaml`: Responds to `@claude` mentions to implement fixes via Claude Code
  - `platform-tests-claude-investigate-and-fix.yaml`: Two-phase workflow that uses Claude to investigate failing tests and automatically open issues and PRs
  - `_move_major_tag.yaml`: Manages the floating `@v1` tag, only moving it when the required package version is published
  - `manual_move_major_tag.yaml`: Escape hatch for manual tag management

- **GitHub Actions** (`.github/actions/`):
  - `checkout-restore-dependencies/action.yaml`: Composite action for checkout, Node setup, and dependency caching with npm token isolation

- **Helper Scripts** (`.github/scripts/`):
  - `run-with-apify-tokens.mjs`: Securely passes only required Actor tokens to CLI commands, preventing secret leakage to `node_modules`

- **Configuration**:
  - `.github/workflows-min-package-version`: Declares the minimum CLI version required by the workflows, gating tag movement until that version is published

- **Documentation**:
  - Updated `README.md` with workflow usage examples, secret handling details, and versioning guidance
  - Updated `CONTRIBUTING.md` to document the workflows as a third component of the package

## Notable Implementation Details

- **Secret Isolation**: Secrets are passed only to the steps that need them, never as job-wide environment variables. The `run-with-apify-tokens.mjs` script reads `apify-test-tools.config.json` to determine which Actor tokens to pass, preventing accidental exposure.

- **Versioning Strategy**: Consumer repos pin workflows to `@v1` (a floating major tag), not `@master`. The tag only moves when the declared minimum package version is published on npm, preventing workflows from calling unreleased CLI features.

- **Claude Integration**: The `platform-tests-claude-investigate-and-fix.yaml` workflow chains investigation and fix phases as job dependencies rather than label triggers, working around GitHub's limitation that scheduled runs cannot trigger other workflows.

- **Backward Compatibility**: The `platform-tests.yaml` workflow supports both the deprecated `subtest` input and the new `test-files-glob` input for gradual migration.

https://claude.ai/code/session_01SkUZADZ6gzWW4GE6CFMreM


.github/workflows/_move_major_tag.yaml
```diff
@@ -0,0 +1,100 @@
...
+        runs-on: ubuntu-latest
+        permissions:
+            contents: write
+        env:
+            # The major version consumers pin, e.g.
+            # `uses: apify/apify-test-tools/.github/workflows/public_pr-build-test.yaml@v0`.
+            # Bump this only for a breaking change to a workflow's inputs, secrets or behaviour.
+            # The previous tag then stops moving and keeps working, so repos migrate when they get
+            # to it instead of all at once.
+            #
```
ruocco-l, line 39: Not that is the most pressing thing, but migrating the tag from v0 to v1 is confusing for me. It's true that v0 freezes, but the actions will still install the latest version of the library (which will be allowed by the floor because it would set a lower version of the library for sure). That way we will probably have a library that works just for v1 and v0 breaks as soon as the newer version of the library is published.

.github/workflows/public_pr-build-test.yaml
ruocco-l, line 1: @artogahr and @JuanGalilea smuggled two commits in while you were doing this so this action is not 1:1 what we have in the other repo 👀

ruocco-l (commented):
It looks good to me, just a question on the v0->v1 migration, which maybe I don't quite fully understand 😅

.github/workflows/public_pr-build-test.yaml
metalwarrior665: I will sync the new updates before merging

.github/workflows/_move_major_tag.yaml
metalwarrior665: Good catch. Claude instead went with the exact version pin per action release tag. I think that also generally simplifies the mental model, each actions commit release will bound to extract library version.

https://github.com/apify/apify-test-tools/pull/120/changes/277f959f6ab33b50a4b0cff76f5cb29e7c2bc1ad

JuanGalilea:
so the idea is for this repo to become all things CI/CD? since there are several unrelated actions 👀 (like claude stuff)

metalwarrior665:
> so the idea is for this repo to become all things CI/CD? 

The main reason is that the actions are closely tied to this repo and we need to deploy them in sync. Haing them here allows us to to have the plumbing for that. Also one less repo to maintain.

> since there are several unrelated actions 👀 (like claude stuff)

It is called `apify-test-tools` so I can imagine it can have stuff that you just pick like a buffet. But I could imagine that we host some unrelated actions separately. For now it will be easier to have it in one repo though.

JuanGalilea (commented):
can we move the actions not being used in this repo into a subfolder?
So at least we know which ones are used and which ones are "exports"?
I know the cost of this will be one more `./something` when calling them but some order here would be benefitial

JuanGalilea:
btw, this will fuck up sync https://github.com/apify-store/github-actions-source/pull/67 (its very short tho)

metalwarrior665:
Sadly,.Github seems to force workflows to be in the same top level folder so the only solution is a prefix.

JuanGalilea (approved):

PR #122, opened 2026-08-20
refactor: Remove need for RUN_PLATFORM_TESTS env var, rely on TESTER_APIFY_TOKEN

### LINKED ISSUE #121: Remove RUN_PLATFORM_TESTS env var
Instead, check the presence of APIFY_TESTER_TOKEN

We can do this before v1 since this won't break anyone

metalwarrior665:
Closes https://github.com/apify/apify-test-tools/issues/121

Shouldn't be breaking for anyone, only if you would configure `RUN_PLATFORM_TESTS` locally while also passing `TESTER_APIFY_TOKEN`, unlikely

## Summary
Removed the `RUN_PLATFORM_TESTS` environment variable and simplified the platform test gating logic to rely solely on the presence of `TESTER_APIFY_TOKEN`. This reduces configuration complexity while maintaining the same functionality.

## Key Changes
- Removed `RUN_PLATFORM_TESTS` from environment variable destructuring in `lib/lib.ts`
- Updated the `describe` function's `runIf` condition to check `!!TESTER_APIFY_TOKEN || !!RUN_ALL_PLATFORM_TESTS` instead of `!!RUN_PLATFORM_TESTS || !!RUN_ALL_PLATFORM_TESTS`
- Added comprehensive documentation explaining when platform tests run and why `RUN_ALL_PLATFORM_TESTS` is needed for periodic tests
- Updated README.md to remove `RUN_PLATFORM_TESTS=1` from all example commands
- Added clarification that platform test suites are automatically skipped unless `TESTER_APIFY_TOKEN` is set, keeping regular unit test runs unaffected

## Implementation Details
The change leverages the fact that `TESTER_APIFY_TOKEN` is already a required dependency for platform tests (needed to instantiate the ApifyClient). By using its presence as the gating condition, we eliminate redundant configuration while maintaining backward compatibility with the `RUN_ALL_PLATFORM_TESTS` flag for periodic test scenarios where no `ACTOR_BUILDS` environment variable is available.

https://claude.ai/code/session_01JSd2mJ6ZFg8cf7dRg4nTQg

### DIFF
--- a/lib/lib.ts
+++ b/lib/lib.ts
@@ -43,6 +43,9 @@ const DEFAULT_TEST_OPTIONS: ActorTestOptions = {
 
 /**
  * Platform tests only run when `TESTER_APIFY_TOKEN` is provided, since they need the platform to run against.
+ *
+ * `RUN_ALL_PLATFORM_TESTS` is needed for periodic tests, where there is no `ACTOR_BUILDS` env var to match
+ * the tests against - without it, every `testActor` would be filtered out as an actor we didn't build.
  */
 export const describe = (name: string, fn?: SuiteFactory<object>, options: ActorTestOptions = DEFAULT_TEST_OPTIONS) => {
     vitestDescribe.runIf(!!TESTER_APIFY_TOKEN || !!RUN_ALL_PLATFORM_TESTS)(name, options, fn);
--- a/lib/lib.ts
+++ b/lib/lib.ts
@@ -42,10 +42,10 @@ const DEFAULT_TEST_OPTIONS: ActorTestOptions = {
 };
 
 /**
- * Platform tests only run when `TESTER_APIFY_TOKEN` is provided, since they need the platform to run against.
+ * Platform tests need `TESTER_APIFY_TOKEN` to talk to the platform, so without it we skip them altogether.
  *
- * `RUN_ALL_PLATFORM_TESTS` is needed for periodic tests, where there is no `ACTOR_BUILDS` env var to match
- * the tests against - without it, every `testActor` would be filtered out as an actor we didn't build.
+ * `RUN_ALL_PLATFORM_TESTS` enables them too because locally we can test against a hardcoded `runId`,
+ * which doesn't need the tester token.
  */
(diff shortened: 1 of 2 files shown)


lib/lib.ts
```diff
@@ -41,8 +41,14 @@ const DEFAULT_TEST_OPTIONS: ActorTestOptions = {
     timeout: DEFAULT_TEST_RUN_DURATION_MS,
 };
 
+/**
+ * Platform tests only run when `TESTER_APIFY_TOKEN` is provided, since they need the platform to run against.
+ *
+ * `RUN_ALL_PLATFORM_TESTS` is needed for periodic tests, where there is no `ACTOR_BUILDS` env var to match
+ * the tests against - without it, every `testActor` would be filtered out as an actor we didn't build.
+ */
```
ruocco-l: This comment is a bit misleading: we do need the `RUN_ALL_PLATFORM_TESTS` to run the scheduled tests but we also need the `TESTER_APIFY_TOKEN` for those.

The reason why the `|| !!RUN_ALL_PLATFORM_TESTS` earns its place in this check is because locally we can do test on hardcoded runId, so we don't need the `TESTER_APIFY_TOKEN`.

The comment you wrote should go next to the `shouldRun` const in `testActor` and `testStandbyActor`

ruocco-l (approved):
Just a NIT on a comment

lib/lib.ts
metalwarrior665: Good catch, I updated the code. 

> The reason why the || !!RUN_ALL_PLATFORM_TESTS earns its place in this check is because locally we can do test on hardcoded runId, so we don't need the TESTER_APIFY_TOKEN.

This is really confusing, let's think about how to refactor it. `RUN_ALL_PLATFORM_TESTS` shouldn't be used as a hack to enabled local tests that don't call Actors. Also without `TESTER_APIFY_TOKEN`, some runs might not be public (they generally are on our account but for community they might not be, by default they are not public anymore)

metalwarrior665: https://github.com/apify/apify-test-tools/issues/124

oklinov (approved):

metalwarrior665:
Oops forgot to merge this

PR #126, opened 2026-08-21
refactor: unify build orchestration flows and fix runBuilds dry-run output

ruocco-l:
- Eliminated duplication caused by `startActorBuild/startActorBuildFromLocal` and `runBuilds/runBuildsFromLocal`
- Fixed dryRun branch on build so now only returns `BuildData` (this can maybe be expanded by adding additional debug logs)
- Replaced unnecessary `ts-ignore` inherited by old `apify-client` version

### DIFF
--- a/bin/build.ts
+++ b/bin/build.ts
@@ -12,6 +12,14 @@ type BuildPrActorOptions = {
     actorConfig: ActorConfig;
     useDockerCache: boolean;
 };
+
+// Fixed version number used to build from local source files, since there is no real version to track.
+export const ZIP_VERSION = '0.98';
+
+type ActorClient = ReturnType<ApifyClient['actor']>;
+// NOTE: I couldn't find this type, so I had to extract it :(
+type ActorVersion = Parameters<ReturnType<ActorClient['version']>['update']>[0];
+
 export class ApifyBuilder {
     private constructor(
         private readonly apifyClient: ApifyClient,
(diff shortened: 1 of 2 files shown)


bin/build.ts
```diff
@@ -12,6 +10,10 @@ type BuildPrActorOptions = {
     actorConfig: ActorConfig;
     useDockerCache: boolean;
 };
+
+// Fixed version number used to build from local source files, since there is no real version to track.
+export const ZIP_VERSION = '0.98';
```
metalwarrior665:
```suggestion
export const LOCAL_SOURCE_VERSION_NUM = '0.98';
```

bin/build.ts
```diff
@@ -12,6 +10,10 @@ type BuildPrActorOptions = {
...
+
```
metalwarrior665, line 17: Since we are hoisting the version here, I would also move the 0.99 as `TESTING_VERSION_NUM`

bin/build.ts
```diff
@@ -95,43 +85,34 @@ export class ApifyBuilder {
...
-        const actorClient = this.apifyClient.actor(this.actorFullName);
-        const actorInfo = await actorClient.get();
-        if (!actorInfo) {
-            throw new Error(
-                `No actor named '${this.actorFullName}' was found on the platform. If this` +
-                    ' is unexpected, make sure the actor you are targeting is spelled the' +
-                    ' same as the folder in the repository.',
-            );
-        }
+    startActorBuild = async ({
```
metalwarrior665: There are so many similarly named functions which is really hard ot read, I would prefer to inline this and `startActorBuildFromSourceFiles`. It really just creates the version object and calls `createVersionAndBuild`. I think it would be cleaner that way.

metalwarrior665 (changes requested):
I think it is for the better but few small notes

metalwarrior665 (approved):

PR #129, opened 2026-09-11
feat: allow config file to have config globs

### LINKED ISSUE #128: Create globs for config file
`apify-test-tools.config.json` requires every actor entry to repeat shared fields (e.g. `tokenEnvVar`) individually, even when the value is identical across most/all actors. This gets worse as we add more shared per-actor config (e.g. notification configs).
 
## Solution

Add an `overrides` array that applies shared field values to actors by glob match, instead of repeating them per entry. The concrete `actors` array (literal `folder` + `actorFullName`) is unchanged.

**Design decisions:**
- Two independent glob types: `folder` glob and `actorFullName` glob. `*` = all actors (just a glob with no literal segments, nothing special-cased).
- Precedence (highest to lowest):
  1. Literal value on the actor entry itself (in `actors` array) — never overwritten by an override.
  2. `actorFullName` glob override.
  3. `folder` glob override.
  4. Within the same glob type, last matching entry in the array wins.

ruocco-l:
Closes #128

- New optional `configs?: ActorGlobConfigEntry[]` on the config file. Each entry is `{ match, set }`: `match` requires at least one of `folder`/`actorFullName` (glob patterns via `minimatch`), and both can be set together as an AND match (more specific than either alone). `set` is required — a `match` with nothing to apply is a misconfiguration and `readConfigFile` throws.
- The merge is **field-agnostic** — `set` overlays whatever properties it carries with no special knowledge of `tokenEnvVar` or any other field name. An unrecognized/misspelled property merges in and simply has no effect, same as today.
- Precedence, lowest to highest: matching `folder`-only entries → matching `actorFullName`-only entries → matching combined (`folder` AND `actorFullName`) entries → the actor's own literal entry (never overridden). Within a single tier, only the _last_ matching entry (array position, not pattern specificity) applies — earlier same-tier matches are dropped outright, not merged in.
- Across tiers, results **deep-merge** instead of shallow-overwriting: a higher tier wins on any field it sets, but a field it doesn't set is inherited from a lower tier rather than being lost. Nested objects merge key by key (so a future field like `notifier: { slack: {...}, email: {...} }` can be filled in piecemeal across tiers instead of one config entry clobbering the whole object), and array fields (e.g. `overrideActorContext`) keep the higher tier's array intact and append only the lower tier's genuinely novel entries, preserving the higher tier's order. This is backed by a new hand-rolled, unit-tested `deepMerge` helper.
- Matching follows standard `minimatch` glob rules, see https://github.com/isaacs/minimatch for the full pattern syntax. Notably, "match every actor" needs `**`, not `*`, since `*` doesn't cross path segments.
- `tokenEnvVar` is optional on an actor entry now, but still required overall: once an actor's full config is resolved (literal + any matching `configs`), a `tokenEnvVar` must be present or `readConfigFile` throws — this is the only place the code cares about that specific field name. This is a design choice and should be work as example. Merging the globs doesn't assume anything about the structure of the ending result of the config file, check on the correct shape of the actor configuration must be placed after the globs merge.
- `configs` is entirely optional; a config file without it behaves exactly as before.
- Types: `ActorIdentity` (`folder`/`actorFullName`) and `ActorSettings` (`tokenEnvVar`/`overrideActorContext`) are now shared building blocks — an `actors[]` entry is `ActorDeclaration = ActorIdentity & ActorSettings`, and the resolved output `ActorConfig extends ActorIdentity`, so the two can't silently drift apart. The top-level config file shape is renamed `ConfigFileSchema` (previously `ActorConfigFile`, which read too similarly to the unrelated `ActorConfig`).


bin/utils.ts
```diff
@@ -114,6 +115,62 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
     return undefined;
 };
 
+const validateGlobConfigEntries = (configs: unknown[]): ActorGlobConfigEntry[] => {
+    for (const [index, configEntry] of configs.entries()) {
+        const { folder, actorFullName } = configEntry as ActorGlobConfigEntry;
+
+        // TODO: Allow for combined filtering?
```
metalwarrior665: I would allow this, don't think it is that confusing or dangerous. You simply filter once per folder and then per name. There might be legit use-cases for this

bin/utils.ts
```diff
@@ -3,12 +3,13 @@ import fs from 'node:fs/promises';
...
 
 import type { ActorVersionSourceFile } from 'apify-client';
+import { minimatch } from 'minimatch';
 
 import { SOURCE_FILE_FORMATS } from '@apify/consts';
 
 import { selectActors } from './actor-filtering.js';
 import { isPathWithinScope } from './path-utils.js';
-import type { ActorConfig, ActorConfigFile } from './types.js';
+import type { ActorConfig, ActorConfigFile, ActorConfigFileEntry, ActorGlobConfigEntry } from './types.js';
```
metalwarrior665: These types don't match very well, one says "File", other doesn't.

btw we already have ActorConfig type which is basically the same thing, we should either unify them or derive one from the other. Are there cases where these will differ? If we are simply merging them then they should not differ. No need to solve that in this PR but sooner rather than later.

bin/utils.ts
```diff
@@ -114,6 +115,62 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
     return undefined;
 };
 
+const validateGlobConfigEntries = (configs: unknown[]): ActorGlobConfigEntry[] => {
+    for (const [index, configEntry] of configs.entries()) {
+        const { folder, actorFullName } = configEntry as ActorGlobConfigEntry;
```
metalwarrior665: Could be zod parse I guess but I don't know if you can get such a nice errors from it, don't have experience

bin/utils.ts
```diff
@@ -114,6 +115,62 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
...
+    const matchingActorFullNameConfigs = configs.filter(
+        (configEntry) =>
+            configEntry.actorFullName !== undefined &&
+            typeof actorEntry.actorFullName === 'string' &&
+            minimatch(actorEntry.actorFullName, configEntry.actorFullName),
+    );
+
+    let overlay: Partial<ActorConfigFileEntry> = {};
+    for (const configEntry of [...matchingFolderConfigs, ...matchingActorFullNameConfigs]) {
+        const { folder: matchedFolder, actorFullName: matchedActorFullName, ...rest } = configEntry;
```
metalwarrior665: Let's think a bit if we shouldn't separate the matching fields `folder`,  `actorFullName` vs the configs, it is a bit weird they are on the same level

bin/utils.ts
```diff
@@ -114,6 +115,62 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
...
+    );
+    const matchingActorFullNameConfigs = configs.filter(
+        (configEntry) =>
+            configEntry.actorFullName !== undefined &&
+            typeof actorEntry.actorFullName === 'string' &&
+            minimatch(actorEntry.actorFullName, configEntry.actorFullName),
+    );
+
+    let overlay: Partial<ActorConfigFileEntry> = {};
+    for (const configEntry of [...matchingFolderConfigs, ...matchingActorFullNameConfigs]) {
```
metalwarrior665: If we want to support having both folder and name with AND logic, this would need to change

bin/utils.ts
```diff
@@ -114,6 +115,62 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
...
+    folder: string,
+    configs: ActorGlobConfigEntry[],
+): ActorConfigFileEntry => {
+    const matchingFolderConfigs = configs.filter(
+        (configEntry) => configEntry.folder !== undefined && minimatch(folder, configEntry.folder),
+    );
+    const matchingActorFullNameConfigs = configs.filter(
+        (configEntry) =>
+            configEntry.actorFullName !== undefined &&
+            typeof actorEntry.actorFullName === 'string' &&
```
metalwarrior665: We already validate this eariler and the type should be string | undefined now, no?

metalwarrior665 (changes requested):
This looks simpler than I thought :) I have just minor remarks, feel free to argue against them. Let's wait for Juan but this seems quite uncontroversial.

ruocco-l:
In this PR I postponed the problem that would pose adding the notifier object. As of now, an object would be completely rewritten and it does not have a partial merging logic (e.g. every notifier would get the same token but each actor needs to get its own slack channel for test report). I gave it some thought and even though it's not necessary as of this PR would be merged, we want to add it soon so it makes sense to do it here, so that the diff is tidy.

@metalwarrior665 I'll do your requested changes first and then I'll try to add the deep object merging logic.
@JuanGalilea I'll make this PR a draft until it's ready with the changes I mentioned above, I'll ping you when it's ready to review again.

bin/utils.ts
ruocco-l: I also don't have much experience with zod but when I looked into it I remember that there was a lot of overhead to make it work. I don't think it's necessary, as long as the inputs here are somewhat limited in number

bin/utils.ts
ruocco-l: https://github.com/apify/apify-test-tools/pull/129/commits/332a4379789eeac482f4b432616e9eddde9f0a07
We didn't, but now we do

README.md
```diff
@@ -44,6 +44,35 @@ Each entry has:
...
+}
+```
+
+Each `configs` entry has a `match` (at least one of `folder`/`actorFullName`, glob patterns via [minimatch](https://github.com/isaacs/minimatch)) and a `set` (the fields to fill in on every actor the match applies to). A `match` with both `folder` and `actorFullName` only applies when both match — it's not a match on either one alone.
+
+Precedence, lowest to highest:
+
+1. entries matching on `folder` alone
+2. entries matching on `actorFullName` alone
+3. entries matching on both `folder` and `actorFullName` together
```
metalwarrior665, line 71: This might be more complicated than just order based

bin/utils.ts
```diff
@@ -117,58 +147,88 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
 
 const validateGlobConfigEntries = (configs: unknown[]): ActorGlobConfigEntry[] => {
     for (const [index, configEntry] of configs.entries()) {
-        const { folder, actorFullName } = configEntry as ActorGlobConfigEntry;
+        const { match, set } = configEntry as ActorGlobConfigEntry;
```
metalwarrior665, line 150: I would give this 2nd thought; maybe it is unnecessarily nested now just for the 2 matcher fields.

README.md
```diff
@@ -44,6 +44,35 @@ Each entry has:
...
+4. the actor's own literal entry in `actors[]`
+
+Within a single tier, if more than one entry matches, only the _last_ one (array order) applies — earlier same-tier matches are dropped entirely, not merged in. Across tiers, results are deep-merged from lowest to highest precedence: a higher tier wins on any field it sets, but a field it doesn't set is inherited from a lower tier rather than being lost. Object-valued fields merge key by key; array-valued fields (e.g. `overrideActorContext`) keep the higher tier's array intact and append only the lower tier's entries that aren't already present, preserving the higher tier's order.
```
metalwarrior665, line 74: This is too complex. I would drop the notion of tiers completely. 1-3 can just be taken in order (any use-case for the current way?). And the point 4 can stay undocumented (for backward compat) and we just append it at the end of the glob array so it will override any conflicting fields before.

> only the _last_ one (array order) applies — earlier same-tier matches are dropped entirely, not merged in

I don't get this. I thought the whole goal is to merge here. E.g. one glob sets token on Actor X, another glob sets slack on Actor X etc., these are merged in the config object. Theoretically, we could only merge the top level config fields but I'm not sure about this

> array-valued fields (e.g. `overrideActorContext`) keep the higher tier's array intact

I think we should not merge arrays. They shouldn't include more nested objects so there won't be need for any deep merging. Users can just retype the whole array in the later glob.

metalwarrior665 (changes requested):

bin/utils.ts
ruocco-l: Actually I quite like the assertiveness of `match` and `set`

metalwarrior665: Sure, let's see what @JuanGalilea thinks

README.md
ruocco-l: Alright, I will just do it by order, it's much easier to understand, you are right, I was just hooked on the "more specific" idea, especially with the filter combo.

 I would concede the arrays overwrite (I don't see a particular use for now, other than avoid typing a specific folder that goes into multiple overrideActorContext) but I would keep the deep merge and not to just top level. You could have:
```json
"configs": [
    { "match": { "folder": "**" }, "set": { "notifiers": { "slack": { "tokenEnvVar": "SLACK_TOKEN" } } } },
    { "match": { "folder": "actors/*" }, "set": { "notifiers": { "slack": { "targets": { "release-report": { "dev": "#releases" } } } } } },
    { "match": { "actorFullName": "myteam/web-scraper" }, "set": { "notifiers": { "slack": { "targets": { "test-report": "#ws-tests" } } } } }
]
```
  and this will allow to avoid resetting the token every time and still give flexibility on the other options.

metalwarrior665: Agree

bin/utils.ts
JuanGalilea: i was thinking of something like `Record<glob, config>`.
ohhhh your matcher can do both folders and names, makes sense actually.

Yeah, `match` and `set` do it for me.

bin/utils.ts
```diff
@@ -114,6 +145,92 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
     return undefined;
 };
 
+const validateGlobConfigEntries = (configs: unknown[]): ActorGlobConfigEntry[] => {
```
JuanGalilea, line 148: this whole function is just one zod schema 👀
It would probably take wayy less mental overhead and probably be parsed better 😅

bin/utils.ts
```diff
@@ -114,6 +145,92 @@ const findOverlappingContextPaths = (contextPaths: string[]): [string, string] |
...
+): Record<string, unknown> | undefined => {
+    const tierMatches = matchingConfigs.filter((configEntry) => isInTier(configEntry.match));
+    return tierMatches.length > 0 ? tierMatches[tierMatches.length - 1].set : undefined;
+};
+
+// Precedence, lowest to highest: matching folder-only entries, matching actorFullName-only entries,
+// matching combined (folder AND actorFullName) entries, the actor's own literal entry. The tier
+// winners deep-merge in that ascending order, so a field one tier doesn't set is inherited from
+// a lower tier instead of being clobbered.
+export const mergeGlobConfigs = (
```
JuanGalilea, line 211: fuck this is complicated, are the tiers even necessary?

Also just had a long chat with some guys from the team and now I'm pretty sure this config scheme is not the way to go 😭.

I'll think a bit more about this today and get back to you

metalwarrior665: We agreed to remove the tiers [here](https://github.com/apify/apify-test-tools/pull/129#discussion_r4014271877). But happy to hear the arguments from the outside.

test/unit/bin/utils/config/parser.test.ts
JuanGalilea: add one test here parsing one with globs, so as to test that the whole thing is doing what its supposed to.
Its slightly redundant but still valuable imo.

I did 1 with grouped somewhere here.

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const SETTING_KEYS = ['tokenEnvVar', 'overrideActorContext'] as const;
+
+const schema = z.object({
+    actors: z
+        .array(
+            z.object({
+                folder: z.string(),
+                actorFullName: z.string().regex(ACTOR_FULL_NAME_REGEX),
+                tokenEnvVar: z.string().optional(),
```
JuanGalilea: THEY CAN HAVE STUFF HERE?
whats the precedence?

Also, nice for `legacy` compat, but slightly worrying

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+        .min(1),
+    configs: z
+        .array(
+            z.object({
+                match: z
+                    .object({
+                        folderGlob: z.string().optional(),
+                        actorFullNameGlob: z.string().optional(),
+                    })
+                    .refine((data) => Object.keys(data).length > 0, 'Invalid input: At least one glob is required.'),
```
JuanGalilea: i hate that this is the zod standard for this kind of thing.

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+// folders are matched as written (no normalization) and dotfolders are skipped
+const matches = (actor: Actor, { folderGlob, actorFullNameGlob }: Match) =>
+    (folderGlob === undefined || minimatch(actor.folder, folderGlob)) &&
+    (actorFullNameGlob === undefined || minimatch(actor.actorFullName, actorFullNameGlob));
+
+export const GLOBS_PARSER = defineStrategy(CONFIG_FILE_STRATEGY.GLOBS, schema, (body: GlobsConfig) => {
+    const errors: string[] = [];
+
+    const resolved = body.actors.flatMap((actor): ResolvedActorConfig[] => {
+        const result = { ...actor };
```
JuanGalilea: `structuredClone` might be safer if we ever get some nesting on the actor fields

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// folders are matched as written (no normalization) and dotfolders are skipped
+const matches = (actor: Actor, { folderGlob, actorFullNameGlob }: Match) =>
+    (folderGlob === undefined || minimatch(actor.folder, folderGlob)) &&
+    (actorFullNameGlob === undefined || minimatch(actor.actorFullName, actorFullNameGlob));
+
+export const GLOBS_PARSER = defineStrategy(CONFIG_FILE_STRATEGY.GLOBS, schema, (body: GlobsConfig) => {
+    const errors: string[] = [];
+
+    const resolved = body.actors.flatMap((actor): ResolvedActorConfig[] => {
```
JuanGalilea: holy function nesting batman!
inline function that inlines a function into flatmap? 👀 
The flatmapped one might benefit of some testing without all the other fluff man

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+        const result = { ...actor };
+        // keep tracks of who set the key, the actor definition or the index of the config
+        const owners = new Map<(typeof SETTING_KEYS)[number], 'actor' | number>();
```
JuanGalilea: smort for debuggability

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+                if (owner === undefined) {
+                    owners.set(key, index);
+                    Object.assign(result, { [key]: config.set[key] });
+                } else if (owner === 'actor') {
+                    errors.push(
+                        `Actor "${actor.actorFullName}": "${key}" is set on the actor and by configs[${index}].`,
+                    );
+                } else {
+                    errors.push(
+                        `Actor "${actor.actorFullName}": "${key}" is set by both configs[${owner}] and configs[${index}].`,
```
JuanGalilea: having configs be an array of rules make these errors very painful to write 😭. Maybe a shortened "selector" view could be nice.

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+                }
+            }
+        }
+
+        const { tokenEnvVar } = result;
+        if (tokenEnvVar === undefined) {
+            errors.push(
+                `Actor "${actor.actorFullName}" has no tokenEnvVar: set it on the actor or through a matching config.`,
+            );
+            return [];
```
JuanGalilea: some Either return could be beneficial here, with the added bonus of no mutation of `error` (not very functional-pilled my man 😂 )

bin/utils/config/structures/globs.ts
JuanGalilea: tracking useless globs (they are not owner for no key) might be beneficial to steer people out of dumb configurations

test/unit/bin/utils/config/structures/globs.test.ts
JuanGalilea: splitting the matching function into a pure function taking `actor` + `globConfigs[]` might be good for testability

bin/utils/config/structures/globs.ts
```diff
@@ -0,0 +1,97 @@
...
+                        actorFullNameGlob: z.string().optional(),
+                    })
+                    .refine((data) => Object.keys(data).length > 0, 'Invalid input: At least one glob is required.'),
+
+                set: z
+                    .object({
+                        tokenEnvVar: z.string().optional(),
+                        overrideActorContext: z.array(z.string()).optional(),
+                    })
+                    .refine((data) => Object.keys(data).length > 0, 'Invalid input: At least one setting is required.'),
```
JuanGalilea: we should do a non-empty object refining function as a zod utility.
This call is strike 3 on the duplication side (i did it once and you have 2 more)

JuanGalilea (changes requested):
code is too inlined for my taste

Also i have some comments on functionalities.

Seems like the groundwork for modes payed off, since it seems we can add these pretty easily

bin/utils/config/structures/globs.ts
ruocco-l: There is no precedence, if the configs try to set something that it is already there, it errors. It's why there is the `owner` that check who set the key

JuanGalilea: yeah, saw it in the implementation. I think its good.

PR #79, opened 2026-04-14
feat: add default 1 hour test run timeout

Patai5:
https://apify.slack.com/archives/C059M9L8ZLP/p1776154991254529
To prevent orphaned actor runs from continuing to run even after the tests have ended.

PR #134, opened 2026-09-14
fix: apply default actor timeout to runs

Patai5:
I was checking some run and noticed that it didn't have any timeout in our tests... So I investigated and found out we were not applying any run timeout to runs in the end.

I've already made a "[fix](https://github.com/apify/apify-test-tools/pull/79)" for this in the past, but it didn't work, because I incorrectly assumed that `DEFAULT_TEST_ACTOR_OPTIONS` are actually for the actor (they were not 😭).

Tested this change locally, and it actually works and applies the timeout this time 🙏 

I've also decided to make the run timeout be one minute less than the actual test timeout, because when the entire test times out, you get messy logs without the run link - so you cannot even match the actor run otherwise...


test/unit/lib.test.ts
```diff
@@ -0,0 +1,29 @@
...
+        vi.spyOn(UtilsModule, 'sleep').mockResolvedValue(undefined);
+
+        const testContext = { task: {}, annotate: vi.fn() as TestContext['annotate'] } as TestContext;
+
+        // Act
+        await createStartRunFn('123', testContext)({ input: {} });
+
+        // Assert
+        expect(actorCallMock).toHaveBeenCalledTimes(1);
+        expect(actorCallMock).toHaveBeenCalledWith({}, { build: undefined, log: null, timeout: 3540 });
```
ruocco-l: Might as well also export the defaults so 3540 is not a magic number

lib/lib.ts
```diff
@@ -285,7 +281,8 @@ const createStartRunFn = <T>(actorId: string, testContext: TestContext) => {
             ...(prefilledInput && (await getActorPrefilledInput(apifyClient, actorId, buildId))),
             ...input,
         };
-        const run = await actor.call(actorInput, { build, log: null, ...options });
+        const actorOptions = { ...DEFAULT_TEST_ACTOR_OPTIONS, build, log: null, ...options };
```
ruocco-l: If you set the test timeout but not the run timeout you could get something like:
```
testTimeout: 10 mins
actorTimeout: 59 mins
```

ruocco-l: Also, this would override the actor timeout every time. Most actors have a 'no-timeout' option so they would gain from the default, but some other genuinely needs a longer timeout. We should respect it (we already have the client, should be an easy call to get the default timeout) and use it as a minimum value

ruocco-l (changes requested):
This would work in most cases (when you let it default with the timeouts), but it kinda breaks when you apply some of them, see the comments 🙏

test/unit/lib.test.ts
Patai5: Good point, the `DEFAULT_TEST_RUN_DURATION_MS` is already exported so we can do the same for the actor run duration as well :)

lib/lib.ts
Patai5: > If you set the test timeout but not the run timeout

I think that's more of a problem of the user if they do so. There is nothing that's stopping users of the library from getting into this situation already.

I only set it to 59 minutes to handle the problems of proper logging in case the test times out before the run finishes. It can just as well be 60 minutes.

We could also have some dynamic logic for either a) logging a warning or b) setting the actor timeout to always 1 minute less or something. But I think that both options are not worth it because 99% of users won't ever change the default timeout of the tests - they might change the actor timeout, and there is nothing stopping them from doing so.

---
> Also, this would override the actor timeout every time. 

Good point for actors that have a timeout less than this. That should be totally respected.
✅ https://github.com/apify/apify-test-tools/commit/ccc5bbf06c75245a88c43f1a9d6d8b1be6f7b073

For actors that have it more than an hour or even unlimited - I don't agree that should respect those. It doesn't make sense if the library is by default set to 1 hour only, and what actors are even these? I don't think we have a single actor in store that needs to run for longer than an hour inside a test 🤔

test/unit/lib.test.ts
Patai5: ✅ https://github.com/apify/apify-test-tools/commit/ccc5bbf06c75245a88c43f1a9d6d8b1be6f7b073

lib/lib.ts
metalwarrior665: There are daily tests that sometimes run over 1h. But I think it is ok to have 1h default for all tests and abort the run at 59 minutes (timeout - 1 min).

JuanGalilea: I would go for an even lower default 😅, 1h is still kind of a lot

ruocco-l: I think 1 hour as a default is fine because if we do it the other way around (set a lower one that will quickly tell you that you that something is wrong) you will just end up overriding it everywhere with a longer time out. This way you would do that only if you really want to the test to run for a very short/long time (which is somewhat nice I think)

Patai5: Sounds like we reached a conclusion 🥳 

Just give me an approval unless you have any remarks and this is good to merge :)

ruocco-l: > For actors that have it more than an hour or even unlimited - I don't agree that should respect those. It doesn't make sense if the library is by default set to 1 hour only, and what actors are even these?

But this is not respecting the timeout then, no? if there is a default on the actor, we have to assume that is there for some reason. I think the default should only be in place of what is not there, in this case the unlimited timeout on the actor.

This would be another instance of the actor timeout being longer than the test timeout. I think that we should at lest warn that this is the case

Patai5: For a lot of actors, we have a default of 1 day (24 hours). And there isn't any real thought behind it other than if things go really south, just please stop...

ruocco-l: Hm ok, that's a good argument. Yeah, I guess it's better to safe guard broken tests that would run forever other than somewhat displeasing a user that needs to run a long test and forgot to set the timeout properly.

ruocco-l (approved):

Issue #106, opened 2026-07-20
Parse actor.json to look for actor-specific files


ruocco-l:
See [PR Comment](https://github.com/apify/apify-test-tools/pull/96#discussion_r3552246484)

We can create a list of files that relate to the actor by parsing the actor.json and classify them as actor specific files (README, changelog and schemas) in `classifyFileChange`

JuanGalilea:
time to actually go at this (again), although maybe solving paths is a prerequisite for this since `actor.json` is just a bunch of fucking paths

PR #135, opened 2026-09-14
Feat/read actor json

JuanGalilea:
sort of refactor of the read flow.

Now reading the config also reads actor.json and adds it to the config for each one.
During that process all paths are also resolved so they can be just `readFile(whatever)` instead of having to deal and think about how they behave relative to one another.

Added a custom config file to `readConfigFile` but this is for ease of testing (which is still missing).

Also added validation to files so we know for sure that they are correct and also what they may not have.

Structure is somewhat final (unless you disagree), so just roast the `bin/utils/*` files as much as you want.

TO-DO:
- [ ] fix utils tests
- [ ] fix ts errors on other tests
- [ ] add tests for `actor-config.ts`


bin/utils/actor-config.ts
```diff
@@ -0,0 +1,119 @@
...
+
+import type { ActorConfig } from '../types.js';
+import { type ActorJson, getActorJsonPath, parseActorJsonAndResolvePaths } from './actor-json.js';
+import { safeReadJsonObjectFile } from './files.js';
+
+export const DEFAULT_CONFIG_FILE_PATH = 'apify-test-tools.config.json';
+
+// #region schema
+
+const TEST_UTILS_ACTOR_SCHEMA = z.object({
```
metalwarrior665, line 15: These schemas should sync with https://github.com/apify/apify-test-tools/pull/129, I like the way it is done there

bin/utils/actor-json.ts
```diff
@@ -0,0 +1,123 @@
...
+                message: 'Expected a `default` dataset',
+            })
+            .optional(),
+    })
+    // `dataset` and `datasets` are mutually exclusive
+    .refine((storages) => !(storages.dataset !== undefined && storages.datasets !== undefined), {
+        message: 'Expected only one of `dataset` or `datasets` to be set',
+    });
+
+export const ACTOR_JSON_SCHEMA = z.object({
```
metalwarrior665, line 34: I don't see much reasons to parse the whole actor.json and also run some validations on it. The platform already validates it during build and it could get out of sync with us. We will just need a few fields that we can simply type

metalwarrior665 (commented):
Let's debate once https://github.com/apify/apify-test-tools/pull/129 is merged

bin/utils/actor-json.ts
JuanGalilea: yeah, we can reduce it. This is just a copy of the JsonSchema then turned into zod by clankers.
This one made me think tho, we might benefit of something like this being exposed as a library for our tools, since we read the actor.json in like 10 different places, and all of them have a different implementation of this.

metalwarrior665: I think it is probably public if you ask Claude around apify-core

JuanGalilea: yes, but not in a type-meaningful way :sob:. Which a zod schema very much is

JuanGalilea:
Since we are not ready with #129 but we agreed that whatever the solution is it will be on a shape of `SomeConfig -> DenormalizedConfig -> ConfigWithActorJson`. (with `DenormalizedConfig` being something like the current config)

I'll split the loading so whatever we decide on for the `SomeConfig -> DenormalizedConfig` can be just plugged in here and it we'll be good to go.

That also seems to line up with what we talked about @ruocco-l. Will turn back to draft until i have that ready to be reviewed.

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,124 @@
+import { dirname, join, normalize } from 'node:path';
```
metalwarrior665: How about the folder being `bin/config-parsers/actor-config.ts`? Potentially also `bin/parsers/configs/actor-config.ts` but too nested for me

metalwarrior665:
This can be closed now?

JuanGalilea:
i guess yeah, at this point it has drifted a bit and since i'll do #147 before #106 it seems it will drift even more

PR #136, opened 2026-09-14
fix: sync configured environment variables to Actor versions

### LINKED ISSUE #110: Support syncing env variables from GH
As of now we can set the secrets once on the version and they will not be overridden when we do the update.

We can add an option to sync env variable from Github secrets/variables so that if we need to change them, we can do it in Github and we don't  have to do it on all miniactors separately, everything will be coordinated by Github.

To do this we just need to update the `actorVersion` we pass when doing `actorClient.version(versionNumber).update(actorVersion)` by adding the `envVars` object, more [information in docs](https://docs.apify.com/api/v2/actor-version-post).

NotVadusha:
Closes #110

Adds optional `envVars` mappings to sync values from the build environment to Actor versions before building. Production builds sync all configured entries; test and local builds sync only those marked `isShared`. Existing unrelated variables are preserved, and missing values fail before version updates.

Forwarding GitHub secrets/variables through the shared workflow requires a separate change in `apify-store/github-actions-source`.

Verified: lint, type-check, build, and tests (162 passed, 2 skipped). Live Actor syncing was not tested.

The docs folder was left intentionally, so once the team has a vision for this, the plan can be updated, and an agent may just align the PR with the expected behaviour


NotVadusha:
This fix was implemented during one of the interview steps, with @metalwarrior665

metalwarrior665:
Thanks a lot! As I mentioned, I think this is fairly close to what we want but we will talk with the team first. Thanks again.

PR #139, opened 2026-09-17
refactor(readConfigFile): rework config file reading and testing

JuanGalilea:
PR number one of the setup for config file reworks.

Featuring:
- Tons of fixtures
- clear cut point for new config file parsing
- rework of `readConfigFile` passing previous mock-based test battery and new fixture-based one
- some config fixes like .vscode and tsconfig misconfigurations


bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+    const rawActors = Array.isArray(contents.actors) ? contents.actors : [];
+    const messages = new Set(parsed.error.issues.map((issue) => describeIssue(issue, rawActors)));
+    throw new Error([...messages].join('\n'));
+};
+
+/**
+ * Stage 3 — normalizes the paths the user wrote and rejects entries that collide once normalized.
+ * The repo root is "" from here on, however it was spelled in the file.
+ */
+export const denormalizeConfig = (config: ActorConfigFile): DenormalizedActor[] => {
```
ruocco-l, line 157: Why is it called denormalize if it normalizes (according to the JSDoc)?

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+    const seenFolders = new Set<string>();
+
+    return config.actors.map((entry) => {
+        const folder = entry.folder === '.' ? '' : stripTrailingSlash(entry.folder);
```
ruocco-l, line 161: I know that this was already here, but i realized that you can trick this by using "./" in the config file. Can you fix it now, since we are touching on it?

test/unit/bin/utils/actor-config.test.ts
```diff
@@ -1,36 +1,102 @@
...
+                folder: 'actors/shopify',
+                actorFullName: 'myteam/shopify',
+                tokenEnvVar: 'APIFY_TOKEN',
+                overrideActorContext: ['packages'],
+            },
+        ]);
     });
-};
+
+    it('loadActorConfig resolves an entry that never came from a config file', async () => {
```
ruocco-l, line 82: I don't understand what this is supposed to mean

ruocco-l (approved):
Just some questions, but looks ok to me, thanks!

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+});
+
+export type ActorConfigFile = z.infer<typeof CONFIG_FILE_SCHEMA>;
+
+/**
+ * The single shape stage 4 understands: an actor entry whose paths have been normalized, with no
+ * trace left of which config flavour produced it. Whatever replaces or extends
+ * {@link parseConfigFile} only has to produce this.
+ */
+export interface DenormalizedActor {
```
metalwarrior665, line 45: 1. I think the naming will need some work. One is Actor and other is ActorConfig, so shouldn't this be `DenormalizedActorConfig`?
2. Do we need this ugly term? We could also do something like `ActorConfigBase` -> `ActorConfigComplete` (or similar)? Or perhaps `ActorConfigWithSchemas` as that is what it does for now

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+                return [contextPaths[i], contextPaths[j]];
+            }
+        }
+    }
+    return undefined;
+};
+
+// The repo root is spelled "" internally but "." in a config file, so report it the way a reader
+// would have written it.
+const displayFolder = (folder: string): string => folder || '.';
```
metalwarrior665, line 75: I would inline this or at least define it closer to where used.

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+
+type ConfigIssue = z.ZodError<ActorConfigFile>['issues'][number];
```
metalwarrior665, line 77: Is `issue` an official term? I would think more about `validationError` or at least `validationIssue`

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+/** Stage 1 — the config file as a plain object, or a failure phrased for whoever wrote it. */
+const readConfigFileContents = async (): Promise<Record<string, unknown>> => {
+    const file = await safeReadJsonObjectFile(CONFIG_FILE_NAME);
+    if (file.success) {
+        return file.contents;
+    }
+
+    switch (file.reason) {
+        case 'invalid-json':
+            throw new Error(`Config file "${CONFIG_FILE_NAME}" contains invalid JSON.`);
```
metalwarrior665, line 126: Add a comment we might want to support jsonc in the future

bin/utils/files.ts
```diff
@@ -0,0 +1,57 @@
...
+    failure: new Error(message),
+});
+
+/**
+ * Reads a file and parses it as a JSON object, reporting every way that can go wrong as a value
+ * instead of a thrown error — the caller decides which failures are fatal and how to word them.
+ */
+export async function safeReadJsonObjectFile(path: string): Promise<JsonObjectReadResult> {
+    // `stat`, not `lstat`: a symlink must resolve to its target, matching what `readFile` does below.
+    const stats = await stat(path).catch(() => null);
```
metalwarrior665, line 26: Let's use sync IO, so we don't have to propagate async everywhere. From outside async looks like we will do some network calls.

This is a sequential step in CLI, we don't need wide concurrency like a web server.

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+
+// #endregion
+
+export const readConfigFile = async (selection: { actors: string[]; ignore: string[] }): Promise<ActorConfig[]> => {
+    const rawConfigFile = await readConfigFileContents();
+
+    // replace these two when enabling different file structures
+    // some kind of strategy pattern seems appropriate here
+    const parsedConfigFile = parseConfigFile(rawConfigFile);
+    const entries = denormalizeConfig(parsedConfigFile);
```
metalwarrior665, line 242: Again the variables don't follow. parsedConfigFile -> entries -> actorConfigs. They should derive from one another.

bin/utils/files.ts
```diff
@@ -0,0 +1,57 @@
+import { readFile, stat } from 'node:fs/promises';
```
metalwarrior665, line 1: `files.ts` is too generic. How about `json-object.ts`. We can make more files for other things later

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
+import path from 'node:path';
```
metalwarrior665, line 1: Instead of having a `utils` dump, we could just have `bin/config-parsers/`. We will also add the actor.json parsing to it is already 3 related files.

bin/utils/actor-config.ts
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// #endregion
+
+// #region utils
+
+// Strips a trailing slash so config-declared paths ("actors/shopify/" vs "actors/shopify") compare equal.
+const stripTrailingSlash = (pathValue: string): string => pathValue.replace(/\/+$/, '');
+
+const findOverlappingContextPaths = (contextPaths: string[]): [string, string] | undefined => {
```
metalwarrior665, line 59: Next time it would be good to add PR comment that this function was moved as is so I don't have to read it. Not sure if there is some easy hack for me to do this...

metalwarrior665 (changes requested):
The flow is nice, seems like something we could stick to long term. Just few polishing suggestions.

bin/utils/actor-config.ts
JuanGalilea: it normalizes the paths but the action as a whole is denormalizing (tbh i didn't read the jsdoc my bad my bad).
Will fix it

bin/utils/actor-config.ts
JuanGalilea: i'll fix it in the path resolution pr (PR number 3)
Since there is quite a bit of dubious stuff around paths

bin/utils/actor-config.ts
JuanGalilea: 1. true
2. i mean it is the right term. Maybe `ResolvedActorConfig` or even `ResolvedConfigActor` so it the `ResolvedActorConfig = ResolvedConfigActor[]`?

bin/utils/files.ts
JuanGalilea: YES!!! i did not touch it due to chesterton's fence. but it bugged me the whole time

bin/utils/actor-config.ts
JuanGalilea: i was about to do this now since i started doing groundwork for the multiple config structures and noticed it as an issue.
Will solve it here.

bin/utils/actor-config.ts
JuanGalilea: sometimes github detects it, but its really hit or miss

bin/utils/actor-config.ts
metalwarrior665: sounds good

bin/utils/actor-config.ts
JuanGalilea: set to `ConfigParsingIssue` since its parsing* not validation whats going on

test/unit/bin/utils/actor-config.test.ts
JuanGalilea: me neither, fixed

metalwarrior665 (approved):

